PR #91, opened 2019-11-07
Added hub_tags to specs

jherlitz-r7:
Adds `hub_tags` to spec, same as https://github.com/rapid7/komand-plugins/pull/1524, but in FOSS repo

Will merge this PR into `help-converter-foss` which will be used during the hackathon


abuseipdb/plugin.spec.yaml
```diff
@@ -16,6 +16,10 @@ tags:
 - ip
 - intelligence
 - abuse
+hub_tags:
+  use_cases: []
+  keywords: [ip, intelligence, abuse]
```
jschipp-r7: Should be `key_words` like in the komand-plugins PR for all of these

jschipp-r7:
@jherlitz-r7 will need to update keywords here too

PR #93, opened 2019-11-07
[DF-3527] - Microsoft teams fix auth timeout

jmcadams-r7:
## Proposed Changes

This PR fixes an issue where the New Message Received trigger sometimes failed. 

The exact reason for this failure is unknown, but I suspect, due to this being in Beta, Microsoft may not have implemented authentication in the same way as the v1.0 endpoints. 

To fix that, I added functionality to refresh the API token and reauthenticate if the trigger fails. 

## Testing: 
This was manually tested by running the trigger for 3 hours and watching for token refreshes.  the trigger did hit the reported auth failure, and successfully reauthenticated with the Graph API teams service.


README.md
```diff
@@ -1,8 +1,9 @@
 # InsightConnect Plugins
 
 We made a large number of our plugins open-source in order to benefit our customers, partners, and the greater community.
-The integrations here include some new, some old, and many that need a little TLC. We'll continue to add more over time based on user requests and our own updates.
```
lmilby-r7: Also making changes to the base readme?

jmcadams-r7: No.

lmilby-r7:
@jonschipp Are these readme changes suppose to be here?

jschipp-r7:
> @jonschipp Are these readme changes suppose to be here?

They got merged into `master` yesterday so I think Joey's branch is out of date. Updating and should be fine

microsoft_teams/icon_microsoft_teams/triggers/new_message_received/trigger.py
```diff
@@ -105,10 +105,22 @@ def get_sorted_messages(self, messages_endpoint: str) -> list:
         messages_result = requests.get(messages_endpoint, headers=headers)
         try:
             messages_result.raise_for_status()
+
+
+        # The beta API bombs out every once in a while with Auth denied. Try forcing a refersh of our auth token and
+        # retry getting messages before raising an exception.
         except Exception as e:
```
tschmidtke-r7: Is there a specific exception that is returned when the token expires that could be caught explicitly? Then the refresh and retry would only be run in situations where it is necessary.

jmcadams-r7: No. It's a 400 unauthorized. The beta isn't fleshed out and there are a few things that can cause that. The other issue, if I want to pull the error message, I'll have to have exception handling in my exception handling, and another exception wrapping that. I really don't want to go down the exceptionception path.

tschmidtke-r7: I feel that adding more/better exception handling is preferable to sending a second request unnecessarily.

jschipp-r7: To Tyler's point, I think it's better to ask what is the best, not most convenient, way to solve the problem for the customer. If we can provide better messaging/logging by getting to the core error message then that seems like it's the answer.

jmcadams-r7: I would put better messaging around this if I could. But right now the errors popping out here aren't well defined. Also, we have no clue why this call fails, my bet right now is the Beta is being a beta and not fully fleshed out. 

In the face of ill-defined errors, and unknown causes. I think this the best I can do around this error handling.

PR #95, opened 2019-11-08
Microsoft teams master merge to master

jmcadams-r7:
## Proposed Changes

See PR here: 
https://github.com/rapid7/insightconnect-plugins/pull/93


microsoft_teams/icon_microsoft_teams/triggers/new_message_received/trigger.py
```diff
@@ -105,10 +105,22 @@ def get_sorted_messages(self, messages_endpoint: str) -> list:
...
+            time.sleep(10) # sleep for 10 seconds to make sure we're not killing the API
+            headers = self.connection.get_headers(True) # This will force a refresh of our auth token
+            messages_result = requests.get(messages_endpoint, headers=headers)
+            try:
+                messages_result.raise_for_status()
+            except Exception as e:
+                raise PluginException(cause=f"Could not get messages from Microsoft Graph API."
+                                            f"Get messages result code: {messages_result.status_code}",
+                                      assistance=messages_result.text) from e
+
```
tschmidtke-r7: I think that we should invest the time into adding additional error handling here so that:

1. Users receive more specific messaging regarding what error is received
2. We don't retry when the request if it's not necessary

jmcadams-r7: Unfortunately, Microsoft only returns a 400 - Unauthorized here. The response.text will be a json object with that as one of the text fields, along with other stuff. 

That 400 can be for a variety of reasons, bad API token, bad Refresh token, bad creds, team id changed, etc...

I'd rather generically try to catch everything there, try one reauth attempt, and bail if I can't get messages on the 2nd try. 

If we make it this far in the code and haven't hit an error yet, I have no clue what could be wrong honestly, and the beta error messages aren't very helpful.

tschmidtke-r7: That's fair. Something for us to watch for in the future. Hopefully we'll get some better messaging as the API matures.

microsoft_teams/icon_microsoft_teams/connection/connection.py
```diff
@@ -60,15 +63,15 @@ def get_token(self):
         self.logger.info(f"Authentication was successful, token is: ******************{self.api_token[-5:]}")
         self.logger.info(f"Detected Permissions: {result_json.get('scope')}")
 
-    def check_and_refresh_api_token(self):
+    def check_and_refresh_api_token(self, forceRefreshToken=False):
```
tschmidtke-r7: Can we snake case `forceRefreshToken` for PEP8 compliance?

microsoft_teams/help.md
```diff
@@ -460,6 +460,7 @@ Examples:
 * 1.0.1 - Fix issue where improper exception could be raised
 * 1.1.0 - New trigger New Message Received | New action Send HTML Message
 * 1.2.0 - New actions Add Member to Team, Remove Member from Team, Create Teams Enabled Group, Delete Team, Add Channel to Team, and Remove Channel from Team
+* 1.2.1 - Fix issue where New Message Received trigger could receive an unauthorized error after susstained use
```
mrinehart-r7: `sustained`

PR #329, opened 2020-02-04
Initial commit of screenshot plugin

emartin-merrill-r7:
## Proposed Changes

This is an "unusual" plugin as it implements behavior more normally associated with a service completely internally.  Specifically, it includes an embedded version of chromium allowing it to generate screen shots from the perspective of the orchestrator.

### Description

  - provides PNG bytes of a website screenshot via selenium/chromium

### Testing

Ran tests and internal 

## PR Requirements

Developers, verify you have completed the following items by checking them off:

### Style

Review the [style guide](https://komand.github.io/python/style.html)

- [x] For dependencies, pin [OS package](https://komand.github.io/python/style.html#dockerfile) and [Python package](https://komand.github.io/python/style.html#requirements-txt) versions
- [x] For security, set least privileged account with ``USER nobody`` in the ``Dockerfile`` when possible
- [x] For size, use the [slim SDK images](https://komand.github.io/python/sdk.html#sdk-versions) when possible: ``komand/python-3-37-slim-plugin`` and ``komand/python-3-37-plugin``
- [x] For error handling, use of [PluginException](https://komand.github.io/python/error_handling.html#plugin-exceptions) and [ConnectionTestException](https://komand.github.io/python/error_handling.html#connection-exceptions)
- [ ] For logging, use [self.logger](https://komand.github.io/python/sdk.html#logging)
- [x] For docs, use [changelog style](https://komand.github.io/python/style.html#changelog)
- [ ] For docs, validate markdown with ``make validate`` which calls ``mdl`` to lint ``help.md``

### Functional Checklist
- [ x] Work fully completed
- [ x] Functional
  - [ x] Any new actions/triggers include JSON [test files](https://komand.github.io/python/style.html#tests) in the `tests/` directory created with `icon-plugin run -c sample $action > tests/$action.json`
  - [x ] Tests should all pass unless it's a negative test. Negative tests have a naming convention of `tests/$action_bad.json`
  - [ ] Unsuccessful tests should fail by raising an exception causing the plugin to die and an object should be returned on successful test
  - [ ] Add functioning test results to PR, sanitize any output if necessary
    * Single action/trigger `icon-plugin run -T tests/example.json --debug --jq`
    * All actions/triggers shortcut `icon-plugin run -T all --debug --jq` (use PR format at end)
  - [ ] Add functioning run results to PR, sanitize any output if necessary
    * Single action/trigger `icon-plugin run -R tests/example.json --debug --jq`
[...]

PR #330, opened 2020-02-05
initial commit url_screenshot (originally submitted from a fork)

emartin-merrill-r7:
Original PR: #329

https://github.com/rapid7/insightconnect-plugins/pull/329


url_screenshot/help.md
```diff
@@ -0,0 +1,68 @@
+# Description
+
+Screenshot a URL from the Orchestrator
+
+# Key Features
+
+Generates a Screenshot of a URL using chromium
```
lmilby-r7:
```suggestion
* Generates a Screenshot of a URL using chromium
```

url_screenshot/help.md
```diff
@@ -0,0 +1,68 @@
...
+
+_This plugin does not contain a connection._
+
+## Technical Details
+
+### Actions
+
+#### Screenshot URL
+
+This action is used to grab a Screenshot of a URL from the Orchestrator.
```
lmilby-r7:
```suggestion
This action is used to grab a screenshot of a URL from the Orchestrator.
```

url_screenshot/icon_url_screenshot/actions/screenshot/action.py
```diff
@@ -0,0 +1,59 @@
...
+        chrome_options = Options()
+        chrome_options.add_argument("--headless")
+        chrome_options.add_argument("--window-size=%s" % self.WINDOW_SIZE)
+        chrome_options.add_argument("--no-sandbox")
+        chrome_options.binary_location = self.CHROME_PATH
+
+        url = params.get(Input.URL, "")
+        delay = params.get(Input.DELAY, 0)
+        if not url.lower().startswith('http'):
+            raise PluginException(cause='Input Error:', assistance='URLs need to start with "http"')
```
mrinehart-r7:
```suggestion
            raise PluginException(cause="Input Error:", assistance="URLs need to start with 'http'")
```

mrinehart-r7: Quote style consistency 👍

url_screenshot/icon_url_screenshot/actions/screenshot/action.py
```diff
@@ -0,0 +1,59 @@
...
+        driver = webdriver.Chrome(executable_path=self.CHROMEDRIVER_PATH, options=chrome_options)
+        driver.get(url)
+        time.sleep(delay)
```
mrinehart-r7: Looks like you should be using WebDriverWait https://stackoverflow.com/questions/26566799/wait-until-page-is-loaded-with-selenium-webdriver-for-python

url_screenshot/icon_url_screenshot/actions/screenshot/action.py
```diff
@@ -0,0 +1,59 @@
...
+    def __init__(self):
+        super(self.__class__, self).__init__(
+                name='screenshot',
+                description=Component.DESCRIPTION,
+                input=ScreenshotInput(),
+                output=ScreenshotOutput())
+
+    def run(self, params={}):
+
+        name = f"/tmp/{str(uuid.uuid1())}.png"
```
mrinehart-r7: A cleaner way to work with the file writing/reading in this function would be to utilize the tempfile library bundled with Python. Example:
```
with tempfile.NamedTemporaryFile(suffix=".png") as t:
    # Print the path to the file, for example
       print(t.name)

>>>> /var/folders/7g/64z6m5x92n12p21mspg9d384xwrlwt/T/tmp0bkf6n14
```

No need to write to the filesystem and the file will automatically be removed from memory after the context closes

url_screenshot/icon_url_screenshot/actions/screenshot/action.py
```diff
@@ -0,0 +1,59 @@
...
+        if full_page:
+            body = driver.find_element_by_tag_name('body')
+            element_png = body.screenshot_as_png
+            img_str = base64.b64encode(element_png).decode('ascii')
+            return {Output.SCREENSHOT: img_str}
+        else:
+            driver.save_screenshot(name)
+            driver.close()
+
+            with open(name, "rb") as file:
```
mrinehart-r7: This will slightly change if you use the tempfile implementation

url_screenshot/plugin.spec.yaml
```diff
@@ -0,0 +1,50 @@
+plugin_spec_version: v2
+extension: plugin
+products: [insightconnect]
+name: url_screenshot
+title: Screenshot a URL
+description: Screenshot a URL from the Orchestrator
```
mrinehart-r7:
```suggestion
description: Screenshot a URL
```
I'm not sure if the orchestrator part is necessary, it may be too technical for the description.

url_screenshot/plugin.spec.yaml
```diff
@@ -0,0 +1,50 @@
...
+- utilities
+- url
+hub_tags:
+  use_cases: [data_enrichment]
+  keywords: [utilities, analysis, url]
+  features: []
+actions:
+  screenshot:
+    title: Screenshot URL
+    description: Takes a screenshot of a URL from the Orchestrator
```
mrinehart-r7:
```suggestion
    description: Take a screenshot of a URL
```

url_screenshot/plugin.spec.yaml
```diff
@@ -0,0 +1,50 @@
...
+      url:
+        title: URL
+        type: string
+        description: URL to screenshot
+        required: true
+        order: 1
+      delay:
+        title: Delay
+        type: integer
+        description: Wait x seconds after page load before taking screenshot
```
mrinehart-r7:
```suggestion
        description: Delay (in seconds) after page has loaded before screenshot
```
Thoughts? Just trying to make the description more succinct.

url_screenshot/plugin.spec.yaml
emartin-merrill-r7: I feel it needs to be in there someplace since it is significantly different behavior from other screenshot plugins  where they screenshot from someplace out on the internet.  Do you have any suggestions for better wording?

url_screenshot/icon_url_screenshot/actions/screenshot/action.py
emartin-merrill-r7: Reviewing the stack overflow link, I don't believe so.  The webdriver.wait is for waiting for a specific element to load in response to an ajax call or similar.  The webdriver.get(url) will wait until the page is loaded.

The delay here is meant to be controllable so as to catch the output of scripts which may be triggering at the completion of loading the page.  Some pages look clean when you first land on them and then, 5 seconds later, they look awful.  This delay is meant to catch the later state.

url_screenshot/icon_url_screenshot/actions/screenshot/action.py
emartin-merrill-r7: Nice!  I'll implement

PR #466, opened 2020-04-29
Remove JSON test files from repo

jschipp-r7:
### Description

  - Remove JSON test files because they clutter the repo
  - Update gitignore to avoid JSON test files


.gitignore
```diff
@@ -49,4 +49,8 @@ demo-*.tar.gz
 *.orig
 
 # Ignore test output generated from icon-plugin tool
-**/.output
\ No newline at end of file
+**/.output
+
+# Ignore .json files
```
zyoutz-r7, line 54: Worth ignoring all `**/tests/*` dir files in case someone creates one without the extension?  If we have json files needed for any reason, just individually exclude from ignore in the plugins?

jschipp-r7: We have a few testing scripts in the tests directory that we want to keep. It's a good call out though. Let me think about this for a different PR

PR #473, opened 2020-05-06
Soar 601 pan os plugin commit action default change - VERSION BUMP only

tylerslijboom-r7:
## Proposed Changes

Pull request from just 45 minutes ago missed the version change, this PR is just that version change.

### DIFF
--- a/palo_alto_pan_os/help.md
+++ b/palo_alto_pan_os/help.md
@@ -23,11 +23,11 @@
 
 The connection configuration accepts the following parameters:
 
-|Name|Type|Default|Required|Description|Enum|
-|----|----|-------|--------|-----------|----|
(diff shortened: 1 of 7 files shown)


palo_alto_pan_os/help.md
```diff
@@ -641,6 +641,7 @@ When using the Add External Dynamic List action, a day and time must be chosen e
 
 # Version History
 
+* 1.5.7 - Default value of Commit action updated
```
jmcadams-r7, line 644:
```suggestion
* 1.5.7 - Update to add the default value of Commit action
```

emartin-merrill-r7: it isn't actually adding the default value - it's fixing the default value which was incorrect.

PR #480, opened 2020-05-12
Added MessageGUID to the output of Proofpoint TAP

tylerslijboom-r7:
## Proposed Changes

Added 1 optional field to the "message" output object, "message_guid" that is extracted from the TAP alert.

### Description

See above

### Testing

icon-plugin -R tests/testing.json parses out the field.


proofpoint_tap/plugin.spec.yaml
```diff
@@ -78,6 +78,11 @@ types:
       description: Header reply to
       type: string
       required: false
+    message_guid:
```
mrinehart-r7, line 81: I don't see this used in code anywhere?

tylerslijboom-r7: proofpoint_tap/komand_proofpoint_tap/util/tap_formatter.py  has it added.  The value will be extracted from the HTML just like the other values in that clean_data hash in tap_formatter.py

tylerslijboom-r7:
> There's a Threat-ID field that we need to extract too.

Added and updated the branch.

jschipp-r7:
Action verified as working

PR #498, opened 2020-05-21
Change Dockerfile and SSH key credential type

PhilippBehmer:
I changed the Dockerfile and the SSH key credential type. 
Successfully tested in Testlab.


ssh/help.md
```diff
@@ -109,6 +109,7 @@ _This plugin does not contain any troubleshooting information._
 
 # Version History
 
+* 4.0.0 - Upgrade the Dockerfile and change the SSH key credential type
```
jschipp-r7, line 112:
```suggestion
* 4.0.0 - Upgrade the plugin runtime to `komand/python-3-37-plugin` and run as least-privileged user | Change the SSH key credential type to `credential_secret_key` to skip PEM validation in the product UI
```

PR #648, opened 2020-09-01
[do not merge] add cloud_enabled tag

sengleson-r7:
Adding the cloud-enabled keyword tag to all plugins that are currently cloud-enabled.

This will be a temporary solution so that we can easily display plugins & workflows that may be run on the cloud in the Extension Library.


jschipp-r7 (dismissed):

jschipp-r7:
@sengleson-r7 You'll want to get an engineer to version bump and regenerate these. Plugins won't release without a version bump

zyoutz-r7:
@jschipp-r7 I'm taking care of these and the workflows

zyoutz-r7:
Closing PR as this won't be merged.  Keeping branch around for now

PR #757, opened 2020-12-05
Trend Micro Deep Security - Improve search for IPS rules by CVE list

PhilippBehmer:
I improved the way how we search for IPS rules.
Instead of making a lot of requests I download a list of all rules and match them.

In a test with around 1800 CVEs the run time went down from 7 minutes to 12 seconds.


trendmicro_deepsecurity/icon_trendmicro_deepsecurity/actions/search_rules/action.py
```diff
@@ -16,57 +17,131 @@ def __init__(self):
...
-                                                    verify=self.connection.dsm_verify_ssl)
+        ips_rules = []
+        ips_id = 0
+        step = 5000 # <= 5000
 
-            self.logger.info(f"CVE: {cve}")
-            self.logger.info(f"url: {response.url}")
-            self.logger.info(f"status: {response.status_code}")
-            self.logger.info(f"reason: {response.reason}")
+        while True:
```
jmcadams-r7, line 73: This loop isn't clear to me. 

I see the break below, but I'm not sure what you're incrementing or changing to make that break ever be hit. 

It looks like you're making the same request over and over and pulling the last ID out of it. 

Suggestion: Don't use While True unless you absolutely have to, make the "True" part an actual If statement for the break condition.

jmcadams-r7 (changes requested):

trendmicro_deepsecurity/help.md
```diff
@@ -9,14 +9,13 @@
...
 
 * Trend Micro Deep Security
+* [Create an API key for Deep Security](https://help.deepsecurity.trendmicro.com/api-key.html)
+* Create a new connection and enter the Deep Security Manager URL and API key
 
 # Documentation
 
 ## Setup
 
-* [Create an API key for Deep Security](https://help.deepsecurity.trendmicro.com/api-key.html)
```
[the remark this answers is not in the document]
PhilippBehmer: OK I can move it back to Setup, I moved it up, because the generator deletes this if it is there.

trendmicro_deepsecurity/icon_trendmicro_deepsecurity/actions/search_rules/action.py
PhilippBehmer: Instead of doing one search API request for every CVE, I receive a list of all rules from Deep Security. 
As the API only allows me to receive 5000 rules with one query I request the rules in batches of 5000 until there are no rules left so response_data["intrusionPreventionRules"] will be empty and break the loop.

PhilippBehmer: And as I don't know how many rules there are / or will be in the future I need to use the while true. 
But I'm open to suggestions on how to solve it in a better way.

trendmicro_deepsecurity/help.md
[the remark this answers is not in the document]
jschipp-r7: Yeah, it's a bug in the generator

trendmicro_deepsecurity/help.md
```diff
@@ -25,6 +24,18 @@ The connection configuration accepts the following parameters:
...
+
+```
+{
+  "dsm_api_key": "12345678-ABCD-1234-ABCD-123456789012:ABCDEFGH-1234-ABCD-1234-ABCDEFGHIJKL:12345678901234567890123456789012345678901234",
+  "dsm_url": "https://192.51.100.100:4119",
+  "dsm_verify_ssl": true
+}
+```
+
+
```
jschipp-r7, line 37: minor but can remove the 2 extra newlines

trendmicro_deepsecurity/plugin.spec.yaml
```diff
@@ -57,7 +57,7 @@ actions:
         type: "[]string"
         description: CVEs to protect against
         required: true
-        example: ["CVE-2005-0045","CVE-2014-0160","CVE-2017-0144","CVE-1337-1337"]
+        example: ['CVE-2005-0045','CVE-2014-0160','CVE-2017-0144','CVE-1337-1337']
```
jschipp-r7, line 60: These should be double-quotes. If it generates it oddly in help.md  then update help.md after generation

trendmicro_deepsecurity/icon_trendmicro_deepsecurity/actions/search_rules/action.py
PhilippBehmer: I tries to use
`while response_data["intrusionPreventionRules"]:`
but then I get a "local variable 'response_data' referenced before assignment" error. 

I revert back to the `while True`, but I'm open to other solutions.

trendmicro_deepsecurity/plugin.spec.yaml
PhilippBehmer: I have some issues with the generator. I'll fix it manually and report the error

jmcadams-r7 (approved):

trendmicro_deepsecurity/icon_trendmicro_deepsecurity/actions/search_rules/action.py
mrinehart-r7: I think `while True` is fine in this instance... is there anything else you can add to the else clause to ensure a loop break?

PhilippBehmer: The loop will break as soon as the request fails or there is no data returned.

jschipp-r7 (approved):

PR #870, opened 2021-03-31
[DO NOT MERGE] Unit Test Git Hub Action

jmcadams-r7:
## Proposed Changes

This PR adds the Git Hub Action WF to run Unit Tests on plugins.

### DIFF
--- a/.github/workflows/unit_test.yml
+++ b/.github/workflows/unit_test.yml
@@ -1,0 +1,21 @@
+name: PluginValidator
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.7'
+
+    - name: Validate Plugin(s)
+      run: |
+        python -m pip install --upgrade pip setuptools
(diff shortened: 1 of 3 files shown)


.github/workflows/unit_test.yml
```diff
@@ -0,0 +1,22 @@
...
+    - uses: actions/setup-python@v1
+      with:
+        python-version: '3.7'
+
+    - name: Run Unit Tests on Plugin(s)
+      run: |
+        python -m pip install --upgrade pip setuptools
+        curl -s https://6c7e8e5fb8565900774116a4db40018d623c9f47b815655f:@packagecloud.io/install/repositories/rapid7/insightconnect-internal-python-tooling-unstable/script.python.sh | bash
+        python -m pip install --user icon-integrations-ci --no-cache-dir --no-warn-script-location
+        python -m pip install --user insightconnect_plugin_runtime --no-cache-dir --no-warn-script-location
```
mrinehart-r7, line 21: This may present some issues - not all plugins are on the latest version of the runtime. It may be better to look into setting up venvs for each plugin so that each plugin can determine which version of the runtime it needs to import.

jmcadams-r7:
Stale, closing, the branch is still there if we need it.

PR #933, opened 2021-06-02
IVM Cloud Plugin Update get asset Action and API Call

aholland-r7:
### Description

Added option to return the vulnerabilities associated with an asset to the get_asset action. Also updated the API call function so it no longer returns status codes, only JSON responses.

### Testing

Tests run and passed


rapid7_insightvm_cloud/icon_rapid7_insightvm_cloud/actions/stop_scan/action.py
```diff
@@ -17,20 +17,14 @@ def __init__(self):
...
-            if response[1].get("status") == "Stopped":
-                return {Output.SUCCESS: True}
-            else:
-                return {
-                    Output.SUCCESS: False,
-                    Output.STATUS_CODE: 400,
-                    Output.MESSAGE: f"Failed to stop scan with ID '{scan_id}'. Status of scan is {response[1].get('status')}",
-                }
+        self.connection.ivm_cloud_api.call_api(url, "POST")
+
```
jmcadams-r7, line 21: Is this operation always instant? Do we need a sleep() here or something of that ilk to let the product catch up?

jmcadams-r7 (dismissed):

rapid7_insightvm_cloud/icon_rapid7_insightvm_cloud/actions/asset_search/action.py
```diff
@@ -31,6 +31,6 @@ def run(self, params={}):
         else:
             resources = self.connection.ivm_cloud_api.call_api("assets", "POST", parameters)
 
-        assets = resources[1].get("data")
```
tschmidtke-r7, line 34: good catch!

rapid7_insightvm_cloud/icon_rapid7_insightvm_cloud/actions/stop_scan/action.py
```diff
@@ -17,20 +17,14 @@ def __init__(self):
...
+        response = self.connection.ivm_cloud_api.call_api("scan/" + scan_id, "GET")
+        if response.get("status") == "Stopped":
+            return {Output.SUCCESS: True}
         else:
             return {
                 Output.SUCCESS: False,
-                Output.STATUS_CODE: response[0],
-                Output.MESSAGE: f"Failed to stop scan with ID '{scan_id}'",
+                Output.STATUS_CODE: 400,
```
tschmidtke-r7, line 28: Is it possible that the status could be something other than 400?

rapid7_insightvm_cloud/plugin.spec.yaml
```diff
@@ -1105,6 +1105,12 @@ actions:
         type: string
         example: "abc978-5678-abc-a5a94a1234b8-asset"
         required: true
+      include_vulns:
```
tschmidtke-r7, line 1108: Could you set a default for this?

rapid7_insightvm_cloud/icon_rapid7_insightvm_cloud/actions/stop_scan/action.py
aholland-r7: At this point, yes. All other types of failure will be caught in the API call, this catch is just is just a double check that a scan was actually stopped, because the stop_scan call passes so long as the scan exists, even if the scan has already completed and therefore can't actually be stopped.

rapid7_insightvm_cloud/icon_rapid7_insightvm_cloud/actions/stop_scan/action.py
aholland-r7: I'm honestly not sure if this is instant or not

rapid7_insightvm_cloud/icon_rapid7_insightvm_cloud/actions/stop_scan/action.py
```diff
@@ -17,20 +17,14 @@ def __init__(self):
...
+                Output.MESSAGE: f"Failed to stop scan with ID '{scan_id}'. Status of scan is {response.get('status')}",
```
tschmidtke-r7, line 29: Is the status code included in the json body of the response?

tschmidtke-r7 (commented):
Just a couple comments

rapid7_insightvm_cloud/icon_rapid7_insightvm_cloud/actions/stop_scan/action.py
tschmidtke-r7: This would be a synchronous call right? So it should only return once it's complete.

rapid7_insightvm_cloud/icon_rapid7_insightvm_cloud/actions/stop_scan/action.py
aholland-r7: No, it never gets returned to the actions anymore

rapid7_insightvm_cloud/plugin.spec.yaml
aholland-r7: Yeah, I probably should

rapid7_insightvm_cloud/icon_rapid7_insightvm_cloud/actions/stop_scan/action.py
tschmidtke-r7: Oh sorry 🤦 I misunderstood this.

tschmidtke-r7 (approved):
lgtm!

rapid7_insightvm_cloud/icon_rapid7_insightvm_cloud/actions/stop_scan/action.py
aholland-r7: Yeah, as far as I can tell looking at the API documentation and poking around, it only returns when it's done

rapid7_insightvm_cloud/icon_rapid7_insightvm_cloud/actions/start_scan/action.py
```diff
@@ -25,7 +25,7 @@ def run(self, params={}):
         if ips or hostnames:
             body = _format_body(hostnames, ips)
             resources = self.connection.ivm_cloud_api.call_api("assets", "POST", params, body)
-            extra_ids = resources[1].get("data")
+            extra_ids = resources.get("data")
             for extra_id in extra_ids:
```
mrinehart-r7, line 29: Will `resources` always have a `data` key? If not, then line 29 will fail due to attempted iteration over `None`. A safer way to write line 28 would be `extra_ids = resources.get("data", [])`, that way line 29 will not crash and will just get skipped over due to no elements in the list.

mrinehart-r7 (changes requested):

rapid7_insightvm_cloud/icon_rapid7_insightvm_cloud/actions/stop_scan/action.py
jmcadams-r7: It's two calls though. You're calling once to do the action, then you're immediately calling again to see the status of that action. In my experience, it's a crap shoot (by product) on if the server is ready for that second call.

aholland-r7: That makes sense, I'll add in a sleep to be safe

mrinehart-r7: @aholland-r7 can you verify this with someone from the IVM API team? A 5 second wait to be safe isn't the end of the world, but if it's not necessary then we shouldn't implement it.

mrinehart-r7 (approved):

mberezin-r7 (approved):
Seems legit

mberezin-r7 (approved):
Still seems legit

jmcadams-r7 (approved):

PR #942, opened 2021-06-14
Turn on unit test runner

jmcadams-r7:
## Proposed Changes
This PR turns on the unit test runner and adds a simple unit test to the base64 plugin

## Testing: 
Need all checks to pass in the PR to validate testing.


mberezin-r7 (dismissed):
LGTM. Intentionally ignoring the plugin version validator error as noted on Slack message. This change just adds a unit test and regenerates the plugin.

gusingh-r7 (dismissed):

plugins/base64/unit_test/test_encode.py
```diff
@@ -0,0 +1,15 @@
+from unittest import TestCase
+from komand_base64.actions.encode import Encode
+import logging
+
+
+class TestEncode(TestCase):
```
mrinehart-r7, line 6: I don't think we want to include this file in the merged PR

mrinehart-r7 (changes requested):

plugins/base64/unit_test/test_encode.py
jmcadams-r7: I disagree. It's the plugin we use to start all other plugins (it's a good starter one anyway). So I like having the model for unit tests go out with this PR.

mrinehart-r7 (approved):

gusingh-r7 (approved):

PR #966, opened 2021-07-01
[SOAR-5965] - Jira Unit Test

jmcadams-r7:
## Proposed Changes
This PR adds some unit tests to the Jira plugin


plugins/jira/unit_test/test_create_issue.py
```diff
@@ -0,0 +1,166 @@
...
+class TestCreateIssue(TestCase):
+    def setUp(self):
+        self.test_conn = Connection()
+        self.test_action = CreateIssue()
+
+        test_logger = logging.getLogger("test")
+        self.test_conn.logger = test_logger
+        self.test_action.logger = test_logger
+
+    def test_create_issue(self):
```
mrinehart-r7, line 103: Can you add additional tests for:
- Issue without a type
- Issue without a description
- Totally bizarre project ID (examples: symbols, upper/lower-case mix, all numbers, etc.)

Jira is a heavy hitter plugin and we get a lot of issues for it, so it would be good to flesh out the test cases a bit more even if it means we need to spend more time writing them. Let me know what you think around this.

mrinehart-r7 (changes requested):

plugins/jira/unit_test/test_create_issue.py
jmcadams-r7: Yup, I wasn't done with this PR. I opened it so I could see my committed code and see where I'm at. 

I didn't add [WIP] my bad.

mberezin-r7:
Validator wants a version bump, Dockerfile USER set to `nobody`, and tag `ticket` changed, possibly to `tickets`?

mrinehart-r7 (approved):

mrinehart-r7 (changes requested):
Bump patch version for ConnectionTestException -> PluginException

mberezin-r7 (dismissed):

mberezin-r7 (approved):

mrinehart-r7 (approved):

PR #968, opened 2021-07-07
SOAR-6152 GitIgnore Update

aholland-r7:
### Description

Able to add .json files into the payloads folder of unit_tests without manually updating the .gitignore.

### Testing

No functional changes, but tested to make sure this works, and it does.


mberezin-r7 (approved):
Thanks!

jmcadams-r7 (approved):

PR #971, opened 2021-07-14
SOAR-6189 Prospector Update

aholland-r7:
### Description

Updated prospector (Static Code Workflow) to including pylint and pyflakes, as well as change to only run on changed plugins.

### Testing

No official tests for prospector other than the fact that works to find errors
Not a plugin action, so no need for unit tests

This was tested locally with the following results. 

Passing:
Screen Shot 2021-07-15 at 4 45 37 PM

Failing:
Screen Shot 2021-07-15 at 3 19 36 PM


jmcadams-r7 (approved):

jmcadams-r7 (dismissed):

mberezin-r7 (approved):
LGTM!

jmcadams-r7 (dismissed):

jmcadams-r7 (changes requested):
Needs to work with cortext_xdr changes. Add verbosity to the output (show the diff output so we know what's going to be tested.)

PR #1001, opened 2021-09-08
Add autoupdate Github Action

mrinehart-r7:
Adds a GitHub Action to keep PRs up-to-date with their base branch.


pmara-r7 (approved):
Should work same as the workflows repo as far as I can tell!

mberezin-r7 (approved):

PR #1011, opened 2021-09-21
Update Palo Alto Cortex XDR help.md

jprice-r7:
As per Slack:

Jon Schipp:  18 hours ago
This is great. Nice work guys! One thought about this is customers can’t use tasks yet so we should create standard note in plugins docs that this is for a future feature, or hide that section until we launch Cloud Events (edited) 

Spencer Engleson  15 hours ago
@jdevlin @sdevlin @jprice could we put a very low-lift notice on "Tasks" documentation in the library? We just want users to know this is a new plugin capability that is not yet available in-product. Something along the lines of, "Plugin Tasks are currently in development! This new plugin capability will collect and deliver events to Insight products for search and use in automation workflows."

### DIFF
--- a/plugins/palo_alto_cortex_xdr/help.md
+++ b/plugins/palo_alto_cortex_xdr/help.md
@@ -511,6 +511,8 @@ Example output:
 
 ### Tasks
 
+Plugin Tasks are currently in development! This new plugin capability will collect and deliver events to Insight products for search and use in automation workflows.
+
 #### Monitor Incident Events
 
 This task is used to monitor incident events.


plugins/palo_alto_cortex_xdr/help.md
```diff
@@ -511,6 +511,8 @@ Example output:
 
 ### Tasks
 
+Plugin Tasks are currently in development! This new plugin capability will collect and deliver events to Insight products for search and use in automation workflows.
+
```
mrinehart-r7, line 515: @jprice-r7 my only concerns with adding this to individual plugin documentation (granted this is only for one plugin) is that this text will get blown away easily, and it could be forgotten about and live here even after we release tasks. How big of a lift is it to add something quick to the Extension Library?

jschipp-r7: There will be more plugins to have this - there's a backlog of them. I recommend we auto-generate this line in the icon-plugin tooling. The tooling already creates this section in help.md, I imagine we could hardcode this line for any tasks and re-release

jprice-r7: I should have been more clear in my description; this is already in the Extension Library listing, I opened this PR so the documentation is consistent with what's in the ExtLib. Up to you guys how you want to handle it

mberezin-r7: I think that it is worth merging this PR so that this repo is in sync with what is presented in the Extension Library.

mberezin-r7 (approved):

mrinehart-r7 (approved):

PR #1223, opened 2022-04-26
[SOAR-9079] Fix Update URL List By ID Non-Required Fields

ablakley-r7:
### Description

  - Update the Update URL List By ID action to handle non-required field 'name' being empty. Previously, if left empty, the action would fail.
  - Update the Update URL List By ID action to successfully update URL list by setting 'type' input to required. If 'type' input was not provided previously, the passed URLs would not be appended to or replace the original URL list.

## PR Requirements

#### Unit Tests

#### In-Product Tests

### Style

- [ ] For size, use the [slim SDK images](https://docs.rapid7.com/insightconnect/sdk-guide/#sdk-guide) when possible: ``komand/python-3-37-slim-plugin`` and ``komand/python-3-37-plugin``
- [ ] For docs, validate markdown with ``make validate`` which calls ``mdl`` to lint ``help.md``

### Functional Checklist
- [ ] Work fully completed
- [ ] Functional
  - [ ] Any new actions/triggers include JSON [test files](https://docs.rapid7.com/insightconnect/style-guide/#tests) in the `tests/` directory created with `icon-plugin run -c sample $action > tests/$action.json`
    * Single action/trigger `icon-plugin run -T tests/example.json --debug --jq`
    * All actions/triggers shortcut `icon-plugin run -T all --debug --jq` (use PR format at end)
    * Single action/trigger `icon-plugin run -R tests/example.json --debug --jq`
    * All actions/triggers shortcut `icon-plugin run -R all --debug --jq` (use PR format at end)

### Assessment

1. Run `make validate` and make sure everything passes
2. Run the assessment tool: `icon-plugin run -A -R all -T all`. For single action validation: `icon-plugin run -A -R tests/my_action.json -T tests/my_action.json`
3. Copy (`icon-plugin ... | pbcopy`) and paste the output in **a new post** on this PR

Screenshot 2022-04-26 at 14 15 00
Screenshot 2022-04-26 at 14 10 43

### DIFF
--- a/plugins/netskope/plugin.spec.yaml
+++ b/plugins/netskope/plugin.spec.yaml
@@ -4,7 +4,7 @@ products: [insightconnect]
 name: netskope
 title: Netskope
 description: This plugin allows users to create, read, update, and delete URL lists in their Netskope environment
-version: 2.0.0
+version: 1.0.1
 vendor: rapid7
 support: rapid7
 supported_versions: ["2022-04-26"]
@@ -412,19 +412,19 @@ actions:
         example: ExampleName
       urls:
         title: URLs
-        description: List of URLs
+        description: List of URLs (to update this value, URL List Category input must also be provided)
         type: "[]string"
         required: false
         example: '["https://example.com", "https://example.com"]'
       type:
         title: URL List Category
-        description: Category of URL list
+        description: Category of URL list (to update this value, URLs input must also be provided)
         type: string
         enum:
           - ""
           - exact
           - regex
-        required: true
+        required: false
         example: exact
     output:
       id:
--- a/plugins/netskope/plugin.spec.yaml
+++ b/plugins/netskope/plugin.spec.yaml
@@ -4,10 +4,10 @@ products: [insightconnect]
 name: netskope
 title: Netskope
 description: This plugin allows users to create, read, update, and delete URL lists in their Netskope environment
-version: 1.0.0
+version: 1.0.1
 vendor: rapid7
 support: rapid7
-supported_versions: ["2022-02-11"]
+supported_versions: ["2022-04-26"]
 status: []
 resources:
   source_url: https://github.com/rapid7/insightconnect-plugins/tree/master/plugins/netskope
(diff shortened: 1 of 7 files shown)


cmcnally-r7 (approved):

jkaluza-rapid7 (approved):

plugins/netskope/plugin.spec.yaml
```diff
@@ -424,7 +424,7 @@ actions:
           - ""
           - exact
           - regex
-        required: false
+        required: true
```
igorski-r7, line 427: Why do we need to change that input parameter to `required = true`? Actually, when `type` has been not entered, only name can be updated by that action. `type` and `urls` fields should be filled out both to change URLs. If we want to change name only, we have to put a value for `name` input only. I think the problem exists here mainly because of the platform's UI - I mean the way it works.

In that situation it has to be described that if we want to update type or URLs - both fields need to be filled there (my bad that I didn't described that in help.md) 😞

ablakley-r7: My theory was that `type` being absent negatively affects an update to the URL list, while it being `required = true` did not affect an update to the URL list name. This way a user couldn't accidentally leave it blank and get a false success which should reduce error rates.

I think either way makes sense though.

igorski-r7: Oh, you are right I think. It won't be able to affect name that way 😃  in case it's required to put both `type` and `urls` in code. Sounds great! 😃  In order to have that working properly -I think it's needed to remove a blank value from enum to that `type` input, and set a `default` to that for example to - `exact`.

ablakley-r7: After discussing with yourself and looking into this more, the `type` / `urls` update is bi-directional. If `type` is not provided when updating `urls` then the update will fail. If `urls` is not provided when updating `type` then the update will fail. 

Making `type` required won't solve both problems, but making both `type` and `urls` required will. The problem with this is that neither values are required at all when updating `name`. 

I think the best solution is yours, to update the help.md with instructions for the user.

igorski-r7 (approved):

blaxminarayan-r7 (approved):

PR #1727, opened 2023-04-25
Cybereason output fix

TKuzas-R7:
### The Issue that is being fixed in this PR:
The plugin action get sensor works individually, however the output the api gives does not match the plugin spec yaml file so we cannot pass the output as an input in insight connect.

### NOTE
Validator is failing because the help_output_validator is faulty. there is a ticket made to fix it.

### The updates for this ticket were the following refactored files:
**Files Refactored:**
- Output for plugin in plugin spec file and files effected by regenerating
  
**Testing**
- Insight Connect Test 
image

### DIFF
--- a/plugins/cybereason/help.md
+++ b/plugins/cybereason/help.md
@@ -24,18 +24,15 @@ The connection configuration accepts the following parameters:
 
 |Name|Type|Default|Required|Description|Enum|Example|
 |----|----|-------|--------|-----------|----|-------|
-|credentials|credential_username_password|None|False|Username and password|None|{"username": "user@example.com", "password": "mypassword"}|
-|hostname|string|None|True|Enter the hostname|None|example.com|
+|credentials|credential_username_password|None|False|Username and password|None|{"username": "https://example.com", "password": "mypassword"}|
+|hostname|string|None|True|Enter the hostname|None|https://example.com|
 |port|integer|8443|True|Enter the port|None|8443|
 
 Example input:
 
 ```
 {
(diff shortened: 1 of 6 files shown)


hpwilson-r7 (approved):

cmcnally-r7 (approved):

plugins/cybereason/help.md
```diff
@@ -136,124 +136,139 @@ This action is used to get sensor.
...
-  "policyName": "ED_CUSTOM_POLICY",
-  "deliveryTime": 1627328611652,
-  "policyId": "8c1c93ff-08cb-4b4e-b253-7f5c9d3e08bc",
-  "compliance": true,
-  "groupId": "af617d16-b83e-4739-948d-339ce21b9177",
-  "groupName": "aaa",
-  "groupStickiness": true,
-  "purgedSensors": false,
-  "groupStickinessLabel": "Manual"
+  "$success": true,
```
igorski-r7, line 163: I don't think we need to include that here 😄 
```suggestion
```

plugins/cybereason/help.md
```diff
@@ -136,124 +136,139 @@ This action is used to get sensor.
...
+  "sensor": {
+    "hasMoreResults": false,
```
igorski-r7, line 165: What about the pagination? Do we handle it anywhere? 😄

igorski-r7 (changes requested):

igorski-r7 (approved):

PR #1998, opened 2023-09-27
[PLGN-410] Duo admin task formatting

dlaverty-r7:
### Description

Removed the cleaning and conversion to camel case of the output for the tasks.
Note: Manual updates to the task schema to avoid any formatting of the task output which means that the validation fails. I want to get this version out for testing in staging so will follow up with corrected unit tests after.


rbowden-r7 (approved):

joneill-r7 (approved):
lgtm

igorski-r7 (commented):
I think `plugin.spec.yaml` file needs to be updated as well for type `log`. It determines the task's output and its still in camel case 😄

igorski-r7 (approved):

PR #2012, opened 2023-10-05
[PLGN-423]Updated timestamp handling

dlaverty-r7:
### Description
Update handling of timestamp functionality with tasks:
- Ensure that the auth logs are ordered ascending in timestamp
- Change now time to 2 mins ago instead of 1 minute as suggested should be the case in the Duo docs 
- Split into maintaining three timestamps - one per log type
- Change the timestamp used to the latest log time instead of the max window time
- Backwards compatibility with the older one last collection timestamp method
- Add additional logging for debug purposes. 

NoteL To minimise risk, have tried to keep changes to a minimum. Also updated unit tests will come after in a subsequent PR.  Schema manually updated with ticket already raised around tooling fix required.

### DIFF
--- a/plugins/duo_admin/komand_duo_admin/tasks/monitor_logs/task.py
+++ b/plugins/duo_admin/komand_duo_admin/tasks/monitor_logs/task.py
@@ -11,6 +11,9 @@ from hashlib import sha1
 
 class MonitorLogs(insightconnect_plugin_runtime.Task):
     LAST_COLLECTION_TIMESTAMP = "last_collection_timestamp"
+    ADMIN_LOGS_LAST_LOG_TIMESTAMP = "admin_logs_last_log_timestamp"
+    AUTH_LOGS_LAST_LOG_TIMESTAMP = "auth_logs_last_log_timestamp"
+    TRUST_MONITOR_LAST_LOG_TIMESTAMP = "trust_monitor_last_log_timestamp"
     STATUS_CODE = "status_code"
     ADMIN_LOGS_NEXT_PAGE_PARAMS = "admin_logs_next_page_params"
     AUTH_LOGS_NEXT_PAGE_PARAMS = "auth_logs_next_page_params"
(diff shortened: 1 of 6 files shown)


plugins/duo_admin/komand_duo_admin/tasks/monitor_logs/task.py
```diff
@@ -28,96 +31,155 @@ def __init__(self):
...
-                    state[self.LAST_COLLECTION_TIMESTAMP] = maxtime_in_milliseconds
 
                 previous_trust_monitor_event_hashes = state.get(self.PREVIOUS_TRUST_MONITOR_EVENT_HASHES, [])
                 previous_admin_log_hashes = state.get(self.PREVIOUS_ADMIN_LOG_HASHES, [])
                 previous_auth_log_hashes = state.get(self.PREVIOUS_AUTH_LOG_HASHES, [])
-                new_trust_monitor_event_hashes = []
-                new_admin_log_hashes = []
-                new_auth_log_hashes = []
+
+                new_trust_monitor_event_hashes = new_admin_log_hashes = new_auth_log_hashes = []
```
joneill-r7, line 104: I think this creates a reference between each variable, rather than setting each value to `[]`? I think the state hashes at the end will all equal the same value? should it not be:

```
new_trust_monitor_event_hashes = new_admin_log_hashes = new_auth_log_hashes = [], [], []
```

llaszuk-r7 (approved):

plugins/duo_admin/komand_duo_admin/tasks/monitor_logs/task.py
llaszuk-r7: but change it for a,b,c = [], [], []

igorski-r7: As above 😄 

```
new_trust_monitor_event_hashes, new_admin_log_hashes, new_auth_log_hashes = [], [], []
```

joneill-r7: My bad copying and pasting and then updating 🤦

igorski-r7 (approved):

joneill-r7 (approved):

PR #2225, opened 2024-01-11
[PLGN-646] Automox 3.0.0 release

dlaverty-r7:
### Description

Release PR for updates provided by @ax-tschmidtke.

Fixes the output schemas for actions that use the action_set type. 
The structure of the output did not match the API result due to improper API documentation.

NOTE: Failing plugin validator issue is due to a known (ticketed) issue which means the validator hits an error when no previous connection_version is available.


igorski-r7 (approved):

llaszuk-r7 (approved):

ax-tschmidtke (approved):
Looks good. What's the right way to retain the version history info when running `insight-plugin refresh` if not having it specified in the spec?

dlaverty-r7:
> Looks good. What's the right way to retain the version history info when running `insight-plugin refresh` if not having it specified in the spec?

Hi @ax-tschmidtke , currently its a manual operation. We have a tooling ticket currently open to try and improve this behaviour.

PR #2272, opened 2024-01-25
PLGN-696 Mimecast: Task filter increase to 96 hours. 

joneill-r7:
Temp fix to move Mimecast to 5.3.6 to extend the lifespan of the filtering to 96 hours. 

To do this I checked out Mimecast 5.3.2 commit, and pulled the files from master. Did an API call to 5.3.2 to match the SDK version to not accidentally bump this. 

Unit tests are passing locally:

Screenshot 2024-01-25 at 17 25 58

Logs locally of first run on 5.3.2 showing more logs being filtered:
```
rapid7/Mimecast:5.3.2. Step name: monitor_siem_logs
First run
Number of raw logs returned from Mimecast: 45
Number of returned logs after filtering performed: 11
```

Logs locally of 5.3.5
```
Plugin task beginning execution...
https://usb-api.mimecast.com
rapid7/Mimecast:5.3.5. Step name: monitor_siem_logs
First run...
About to read file name spameventthread_tse_usb01_11_22_00_19ca0_97ce21fa-ace8-4b8f-be74-f9da3b4b1443_20240201.json
About to read file name spameventthread_tse_usb01_71_86_00_19500_8dd4cefa-fc2b-40d7-98ed-daf239329f8c_20240201.json
About to read file name spameventthread_tse_usb01_11_21_00_120b5_8a04a06a-6723-41ae-9606-818291ee5955_20240125.json
About to read file name spameventthread_tse_usb01_37_51_00_1a4ab_320c213a-9568-42ed-aef1-a7b329bca483_20240131.json
About to read file name spameventthread_tse_usb01_74_91_00_169dc_29366074-b5f2-42c7-b852-cdbe2af0d2fa_20240127.json
About to read file name spameventthread_tse_usb01_6_12_00_17d0a_af777eaa-0ae0-408e-897c-dc7d3cef9c8f_20240129.json
About to read file name spameventthread_tse_usb01_8_15_00_137d1_547e0309-de4d-4cdc-876e-010949032d11_20240131.json
About to read file name spameventthread_tse_usb01_75_94_00_1eb6c_23e083f1-ce31-48f6-bfc8-8848b98fe8e7_20240130.json
About to read file name spameventthread_tse_usb01_69_82_00_1f77d_7d23d6da-be31-46ad-836f-473fb9eb8a13_20240131.json
About to read file name spameventthread_tse_usb01_36_48_00_1ba7c_84cab034-3126-4a9d-b613-3f362e0ed574_20240125.json
[...]

PR #2287, opened 2024-02-01
PLGN-696: adjust filter back to 24 hours and pin SDK version

joneill-r7:
### Description

  - Previously we used to filter by a timeframe of 24 hours. this was bumped in the following PRs:
    - #2275 I bumped the cut off period to 108 hours to allow a backfill of data after coming close to the date we needed. 
    - #2272 Original PR that I bumped to 96 hours. 
    - We now want to put this back to the originally timeframe of 24 hours to get everyone on production inline on one version. 
  - I also added a lot of logging to log each file name, which is excessive and ca now be removed. This is possibly causing of memory issue as well when the pods on production. 

## PR Requirements

#### Unit Tests

- [X] ~Unit tests written for any new or updated code~

#### In-Product Tests

Tested task locally with this version vs the one on production with a larger filter period. Inlcuding screenshot below showing how 5.3.6 filters out a lot more for the same calls:

Screenshot 2024-02-01 at 12 03 10

### DIFF
--- a/plugins/mimecast/plugin.spec.yaml
+++ b/plugins/mimecast/plugin.spec.yaml
@@ -5,7 +5,7 @@ name: mimecast
 title: Mimecast
 description: Services for email security, archiving and continuity. Protect, manage
   and archive without compromise
-version: 5.3.5
+version: 5.3.6
 connection_version: 5
 supported_versions: ["Mimecast API 2022-11-07"]
 vendor: rapid7
@@ -13,7 +13,7 @@ support: rapid7
 cloud_ready: true
 sdk:
   type: slim
-  version: 5.3.0
+  version: 5.3.2
   user: nobody
 status: []
 resources:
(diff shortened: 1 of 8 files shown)


plugins/mimecast/plugin.spec.yaml
```diff
@@ -5,15 +5,15 @@ name: mimecast
...
+version: 5.3.6
 connection_version: 5
 supported_versions: ["Mimecast API 2022-11-07"]
 vendor: rapid7
 support: rapid7
 cloud_ready: true
 sdk:
   type: slim
-  version: 5.3.0
+  version: 5.3.2
```
joneill-r7, line 16: Previously tried to pin to 5.3.0 in the last release but missed the docker file so it still pulled SDK version 5.3.2

rbowden-r7 (approved):

dlaverty-r7 (approved):

PR #2642, opened 2024-07-22
[SOAR-17165] Carbon Black Response Synk Vulnerabilities and SDK Bump

rmurray-r7:
### Description

  - SDK bump to 6.0.0
  - cbapi bump to 1.7.10
  - Snyk vulnerabilities fixed by adding - 'validators', 'urllib3', 'requests' and 'certifi'
  - Mocked Unit tests added for 4 List features
  - Plugin Refresh
  
[Ticket](https://rapid7.atlassian.net/browse/SOAR-17165)
  

### Testing

Due to the API, the following actions cannot be tested: 
- List Alerts
- Isolate Sensor
- Get Binary
- Add Feed
- Unisolate Sensor
- Update Alert
- Uninstall Sensor (Unit Test working)

There is no feeds so the delete feeds action is untestable, but it was working when there was feeds available

#### Unit Tests

All created unit tests are passing successfully:
image

- [x] Unit tests written for any new or updated code


plugins/carbon_black_response/unit_test/test_uninstall_sensor.py
```diff
@@ -11,8 +11,6 @@
 post_uninstall = {
     "systemvolume_total_size": "42947571712",
     "emet_telemetry_path": "",
-    "os_environment_display_string": "Windows 10 Server Enterprise (Evaluation), 64-bit",
-    "emet_version": "",
```
cmcnally-r7, line 15: why did we remove this ? I can't remember exactly why, was it not being used?

plugins/carbon_black_response/requirements.txt
```diff
@@ -1,4 +1,8 @@
 # List third-party dependencies here, separated by newlines. 
 # All dependencies must be version-pinned, eg. requests==1.2.0
 # See: https://pip.pypa.io/en/stable/user_guide/#requirements-files
-cbapi==1.6.2
+cbapi==1.7.10
+validators==0.21.0
+urllib3==2.2.2
+requests==2.32.2
+certifi==2024.7.4
```
cmcnally-r7, line 8: why are all the modules being pinned?

plugins/carbon_black_response/plugin.spec.yaml
```diff
@@ -737,6 +774,7 @@ actions:
         description: The lists of alerts
```
cmcnally-r7, line 774: Include a title in every input and output for every action & trigger throughout the plugin

plugins/carbon_black_response/plugin.spec.yaml
```diff
@@ -766,6 +804,7 @@ actions:
         description: The list of processes
         type: '[]process'
```
cmcnally-r7, line 805: Same again, here and the above/below fields. Won't let me comment on them directly but go through all the inputs/outputs in the actions section of the spec file and give them a title like: `title: Processes` to match the `processes:` part 

Only asking because it seems the titles used to exist in all the schema files and now they've disappeared since you refreshed and the titles don't exist

Probably an issue from switching from `icon-plugin` to `insight-plugin`

cmcnally-r7 (changes requested):
Leaving a small request now to let you work on it while I review the rest

plugins/carbon_black_response/requirements.txt
rmurray-r7: To address Snyk vulnerabilities. When running snyk in terminal to test, these were the required modules to be included

plugins/carbon_black_response/unit_test/test_uninstall_sensor.py
rmurray-r7: It wasn't being used I believe?

cmcnally-r7 (approved):

rbowden-r7 (approved):

PR #2692, opened 2024-08-08
[SOAR-17320] Okta Task connection test & SDK Bump

rmurray-r7:
### Description

  - SDK Bump to 6.0.1 & Plugin Refresh
  - Task Connection test added

### Testing

Testing on Postman to ensure a connection cannot be established without the correct input values

  - Successful Connection
image

  - Invalid URL
image

  - Invalid SecretKey
image

- Invalid URL (with valid URL structure)
image


plugins/html/unit_test/temp_html_2_docx.docx
rmurray-r7, line 1: I am unsure how these were created but I had to commit to=he changes to rebase as it wouldn't delete locally 😢

plugins/okta/komand_okta/connection/connection.py
```diff
@@ -29,3 +31,42 @@ def test(self):
...
+        except ConnectionError as error:
+            self.logger.info(error)
+            raise ConnectionTestException(
+                cause="The URL provided in the connection is unreachable. Please ensure you are providing a valid URL when trying to connect.",
+                assistance="Please verify the URL specified is valid and reachable before attempting to connect again.",
+                data="Please ensure the URL specified can be reached and is valid before trying to reconnect.",
+            )
+
+        except ConnectionResetError as error:
+            self.logger.info("Catching egress rules for connection", error)
```
joneill-r7, line 67: I think you've slightly mistyped this logger? Should 'error' be passed into the logging string? Can we check during testing this logs the URL that we failed to connect to as well please

plugins/okta/komand_okta/connection/connection.py
```diff
@@ -29,3 +31,42 @@ def test(self):
...
+            raise ConnectionTestException(
+                cause="Catching egress rules.",
+                assistance="Catching egress rules.",
```
joneill-r7, line 70: I don't think these are surfaced at the minute, but maybe better having something that would be an issue if we raise it?

```
cause="Failed to connect to supplied URL",
assistance="Please contact support if this issue persists"
``

File: plugins/okta/komand_okta/tasks/monitor_logs/schema.py [hunk 3]
```diff
@@ -44,7 +44,7 @@ class MonitorLogsOutput(insightconnect_plugin_runtime.Output):
   "title": "Logs",
   "description": "All system logs within the specified time range",
   "items": {
-    "type": "object"
+    "$ref": {}
```
joneill-r7, line 47: I think we need to keep this schema untouched or it can cause issues when the output changes in the task

plugins/html/unit_test/temp_html_2_docx.docx
joneill-r7: can jump on a call if you're struggling to remove these from the code changes, just let me know!

plugins/okta/komand_okta/connection/connection.py
joneill-r7: Could we update this cause and assistance field? It should save coming back to update the connection test if testing on staging proves we hit this exception and this cause/assistance is bubbled up anywhere

plugins/okta/komand_okta/connection/connection.py
```diff
@@ -29,3 +31,42 @@ def test(self):
...
+        return_message = "The connection test to Okta was unsuccessful \n"
+        try:
+            _ = self.api_client.list_events("")
+
+            message = "The connection test to Okta was successful"
+            self.logger.info(message)
+            return {"success": True}, message
+
+        except ApiException as error:
+
```
joneill-r7, line 46: a bit nit-picky, but could remove these empty lines between each except block?

joneill-r7 (approved):
lgtm!

PR #2750, opened 2024-08-29
[SI-25434] Abnormal Security - CaseID

rmurray-r7:
### Description

A customer is currently experiencing an issue with the `caseId` being returned as an int and not a string (which is what it is meant to be : https://app.swaggerhub.com/apis/abnormal-security/abx/1.4.3#/Cases/get_cases). 

As the `caseId` is not an input from the user (otherwise we could change the datatype) but an output value generated within the action, this is a possible solution to format it to a string when outputted. 

  - Format the `caseId` to a string when outputted from the action


rmurray-r7:
Closing as I accidentally set the base to master

PR #2751, opened 2024-08-29
[SOAR-17793] Abnormal Security - Case ID issue

rmurray-r7:
### Description

A customer is currently experiencing an issue with the `caseId` being returned as an int and not a string, as well as an issue arising when inputting the `caseId`, as it recognises the input as a integer and not string (which is what it is meant to be : https://app.swaggerhub.com/apis/abnormal-security/abx/1.4.3#/Cases/get_cases). 

Proposed change, converts the `caseId` to a string before being inputted and before being returned, to ensure the issue doesn't appear. Provided a custom plugin to the customer and this solution solved their problem.

SI Ticket: https://rapid7.atlassian.net/browse/SI-25434

  - Format the `caseId` to a string when inputted and outputted from both actions
  - SDK Bump to 6.1.3 & Plugin Refresh

### DIFF
--- a/plugins/abnormal_security/icon_abnormal_security/actions/get_case_details/action.py
+++ b/plugins/abnormal_security/icon_abnormal_security/actions/get_case_details/action.py
@@ -14,4 +14,11 @@ class GetCaseDetails(insightconnect_plugin_runtime.Action):
         )
 
     def run(self, params={}):
-        return {Output.CASE_DETAILS: self.connection.api.get_case_details(params.get(Input.CASE_ID))}
+
+        # Solution to convert Case ID to string if it gets returned as an integer
+        case_id = params.get(Input.CASE_ID, "")
+
+        response = self.connection.api.get_case_details(case_id)
+        if isinstance(response.get("caseId"), int):
+            response["caseId"] = str(response.get("caseId"))
+        return {Output.CASE_DETAILS: response}
--- a/plugins/abnormal_security/plugin.spec.yaml
+++ b/plugins/abnormal_security/plugin.spec.yaml
@@ -40,7 +40,7 @@ version_history:
   - '1.1.0 - New actions Get Cases and Get Case Details'
   - '1.0.0 - Initial plugin'
 references:
-  - '[Abnormal Security](https://abnormalsecurity.com/)'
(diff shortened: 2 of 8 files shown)


igorski-r7 (approved):

ablakley-r7 (approved):

plugins/abnormal_security/plugin.spec.yaml
```diff
@@ -19,7 +24,13 @@ resources:
   license_url: https://github.com/rapid7/insightconnect-plugins/blob/master/LICENSE
   vendor_url: https://abnormalsecurity.com/
 enable_cache: true
+key_features:
+  - "This Plugin does not contain any Key Features."
+requirements:
+  - "This Plugin does not contain any Requirements."
+troubleshooting: "There is no troubleshooting for this plugin."
```
cmcnally-r7, line 31: There should be some key features
Remove requirements and troubleshooting from spec if there is none

plugins/abnormal_security/plugin.spec.yaml
```diff
@@ -29,6 +40,8 @@ version_history:
   - '1.0.0 - Initial plugin'
 references:
   - '[Abnormal Security](https://abnormalsecurity.com/)'
+links:
```
cmcnally-r7, line 43: Links should be this, references should point to API docs that are relevant to the plugin actions

cmcnally-r7 (changes requested):

plugins/abnormal_security/icon_abnormal_security/actions/get_case_details/action.py
```diff
@@ -14,4 +14,11 @@ def __init__(self):
         )
 
     def run(self, params={}):
-        return {Output.CASE_DETAILS: self.connection.api.get_case_details(params.get(Input.CASE_ID))}
+
+        # Solution to convert Case ID to string if it gets returned as an integer
+        case_id = params.get(Input.CASE_ID, "")
+
+        response = self.connection.api.get_case_details(case_id)
```
cmcnally-r7, line 21:
```suggestion
        response = self.connection.api.get_case_details(params.get(Input.CASE_ID, ""))
```

PR #2754, opened 2024-08-30
[SOAR-17529] Palo alto Cortex XDR Task

cmcnally-r7:
### Description

Task implementation for the `monitor_alerts` task. The task connection test and unit tests (started) has also been added and testing outputs have been provided below.

  - Task Implementation for `monitor_alerts` 
  - Task Connection Test added
  - Unit tests

### `PluginValidator` is failing 
The validator is failing because the `connection_version` is bumping from 2 -> 4 and the validator checks for +1 increments. We can ignore this in the context of this plugin and raise a ticket to fix this in the validators

#### Task
Called the Task locally and it pulls the first 100 logs successfully whilst also returning `has_more_pages` as true.

image

#### Task Connection Test

I tested by hitting the endpoint locally via Postman requests to ensure the correct exceptions show when inserting invalid data, and that when hitting the endpoint successfully the correct message shows.

- Successful test:
image

- Invalid URL
image

- Invalid Secret Key/ Key ID
image


plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+        custom_config = {}
+        state = {LAST_SEARCH_FROM: None, LAST_SEARCH_TO: None}
+
+        try:
+
+            now_time = self._get_current_time()
+            now = now_time - timedelta(minutes=15)
+            # last 15 minutes
+            end_time = now.strftime(TIME_FORMAT)
+            print(f"{end_time = }")
```
ablakley-r7, line 69: Looks like a leftover print

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+
+            self.logger.info("Starting to download alerts...")
+
+            response, state, has_more_pages = self.get_alerts_palo_alto(state=state, custom_config=custom_config)
+
+            # TODO - Are we supposed to show the total (MAX LIMIT) or the total (total_count)
+            self.logger.info(f"Total alerts returned: {len(response)}")
+
+            # Local Debugging
+            self.logger.info(f"{type(response) = }")
```
ablakley-r7, line 80: Some leftover log statements.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+
+            # Note - when running locally, we should only get the first 100
+            # When has more pages is True, the task will be rerun automatically again (on staging / the cloud exec)
+
+            # If has more pages == True, then we need to change the search index from 0-100 to 101-200 (Done below now I think)
+            return response, state, has_more_pages, 200, None
+
+        except PluginException as error:
+            self.logger.error(
+                f"A PluginException has occurred. Status code 500 returned. Error: {error.cause}. "
```
ablakley-r7, line 95: I think the status code should be retrieved from response unless the status_code results in an unwanted reaction from PIF. Check out our new wiki page on error handling status codes in PIF integrations.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+        endpoint = "/public_api/v1/alerts/get_alerts"
+        response_alerts_field = "alerts"
+        time_sort_field = "creation_time"
+
+        self.logger.info(f"{ALERT_LIMIT = }")
+
+        # IT GOES 0 -> 100 | 101 -> 200 | 201 -> 300
+        # IF NULL, SET FROM = 0
+
+        if not state.get(LAST_SEARCH_FROM):
```
ablakley-r7, line 117:
```suggestion
        search_from = (state.get(LAST_SEARCH_TO) + 1) if state.get(LAST_SEARCH_FROM) else 0
```

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+        self.logger.info(f"{post_body = }")
+        self.logger.info(f"{url = }")
+
+        try:
+            request = requests.Request(method="post", url=url, headers=headers, json=post_body)
+            response = make_request(
+                request, exception_custom_configs=custom_config, timeout=120, allowed_status_codes=[]
+            )
+
+            response = response.json()
```
ekelly-r7, line 164: Would be better to use `extract_json` from the SDK helper as it has built in exception handling

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+
+        if not state.get(LAST_SEARCH_FROM):
+            search_from = 0
+        else:
+            # STARTS 1 RECORD AFTER THE LAST ONE PULLED
+            search_from = state.get(LAST_SEARCH_TO) + 1
+            self.logger.info(f"{search_from = }")
+
+        # IF NULL SET TO = 100
+        if not state.get(LAST_SEARCH_TO):
```
ablakley-r7, line 125:
```suggestion
        search_to = (search_from + 99) if state.get(LAST_SEARCH_TO) else 100
```

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_incident_events/task.py
```diff
@@ -57,7 +57,7 @@ def remove_old_and_sort_incident_events(
         cleaned_events_after_cutoff.sort(key=lambda x: x.get(time_sorting_field, -1))
```
ekelly-r7, line 57: Not sure if this `monitor_incident_events` task is being used, should it be removed from the plugin?

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+        state[LAST_SEARCH_FROM] = search_from
+        state[LAST_SEARCH_TO] = search_to
+
+        # search_from = state.get(LAST_SEARCH_FROM, 0)
+        # search_to = state.get(LAST_SEARCH_TO, 100) + 100
+
+        # search_to = search_from + ALERT_LIMIT
+        headers = self.connection.xdr_api.get_headers()
+
+        post_body = {
```
ablakley-r7, line 140: Can we put the request into our api.py alongside our other requests? We can separate our logic here so the task.py handles generating the parameters and parsing the response, while the api.py handles the request.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+        self.logger.info(f"{search_from = }")
+        self.logger.info(f"{search_to = }")
+        self.logger.info(f"{headers = }")
+        self.logger.info(f"{post_body = }")
+        self.logger.info(f"{url = }")
+
+        try:
+            request = requests.Request(method="post", url=url, headers=headers, json=post_body)
+            response = make_request(
+                request, exception_custom_configs=custom_config, timeout=120, allowed_status_codes=[]
```
ablakley-r7, line 161:
```suggestion
                request, timeout=120
```

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_incident_events/task.py
cmcnally-r7: Unsure, keeping it in until the end we work out if it is being used or not

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+
+            # If has more pages == True, then we need to change the search index from 0-100 to 101-200 (Done below now I think)
+            return response, state, has_more_pages, 200, None
+
+        except PluginException as error:
+            self.logger.error(
+                f"A PluginException has occurred. Status code 500 returned. Error: {error.cause}. "
+                f"Existing state: {existing_state}"
+            )
+            return [], existing_state, False, 500, PluginException(cause=error.cause, data=error)
```
ekelly-r7, line 98: It doesn't seem like 'failed' status codes are being correctly bubbled up (`500` is hardcoded)

e.g. in the case of invalid credentials, we'd want to return a `401` here, so that the integration goes into a FATAL state

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+
+        # search_from = state.get(LAST_SEARCH_FROM, 0)
+        # search_to = state.get(LAST_SEARCH_TO, 100) + 100
+
+        # search_to = search_from + ALERT_LIMIT
+        headers = self.connection.xdr_api.get_headers()
+
+        post_body = {
+            "request_data": {
+                "search_from": search_from,
```
ablakley-r7, line 142: We need to apply our filter based on creation_time, utilising the gte and lte operators.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+
+        try:
+            request = requests.Request(method="post", url=url, headers=headers, json=post_body)
+            response = make_request(
+                request, exception_custom_configs=custom_config, timeout=120, allowed_status_codes=[]
+            )
+
+            response = response.json()
+
+            total_count = response.get("reply", {}).get("total_count", -1)
```
ablakley-r7, line 166: Just checking, we do we default this to -1? Testing out an empty reply and it should return 0 by default.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+            result_count = response.get("reply", {}).get("result_count", -1)
+            results = response.get("reply", {}).get(response_alerts_field, [])
+
+            is_paginating = ALERT_LIMIT < total_count
+
+            if is_paginating:
+                has_more_pages = True
```
ablakley-r7, line 173: It looks like we set these above as well (lines 131, 132)

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+        )
+
+    def run(self, params={}, state={}, custom_config: dict = {}):  # pylint: disable=unused-argument
+        existing_state = state.copy()
+        has_more_pages = False
+        parameters = {}
+        print(f"{existing_state = }")
+
+        custom_config = {}
+        state = {LAST_SEARCH_FROM: None, LAST_SEARCH_TO: None}
```
ekelly-r7, line 61: Is this for debugging? wondering why the custom_config and state are being set like this

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+                request, exception_custom_configs=custom_config, timeout=120, allowed_status_codes=[]
+            )
+
+            response = response.json()
+
+            total_count = response.get("reply", {}).get("total_count", -1)
+            result_count = response.get("reply", {}).get("result_count", -1)
+            results = response.get("reply", {}).get(response_alerts_field, [])
+
+            is_paginating = ALERT_LIMIT < total_count
```
ablakley-r7, line 170: I think this should be based on whether our result_count is greater or equal to 100 rather than total_count as total_count will always return a static number of results whether we have reached the end or not.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+            self.logger.info(
+                f"{log_msg}Applying the following start time='{state.get(FROM_TIME_FILTER)}'. " f"Limit={alert_limit}"
+            )
+
+        return start_time, alert_limit
+
+    ###########################
+    # Handle Pagination
+    ###########################
+    def _drop_pagination_state(self, state: Dict[str, str]) -> Dict[str, str]:
```
ablakley-r7, line 316: I think we also need to drop the last_search_from and last_search_to to allow to prevent continuous pagination. The API accepts that we can search past the total count.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+
+        post_body = {
+            "request_data": {
+                "search_from": search_from,
+                "search_to": search_to,
+                "sort": {"field": time_sort_field, "keyword": "asc"},
+            }
+        }
+
+        url = urllib.parse.urljoin("https://api-rapid7.xdr.us.paloaltonetworks.com", endpoint)
```
ekelly-r7, line 148: I don't think this url should be hardcoded here...should be coming from the connection or api client instead?

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+                    f"to_time={search_to}"
+                )
+
+            else:
+                has_more_pages = False
+                state = self._drop_pagination_state(state)
+
+            return results, state, has_more_pages
+
+        except PluginException as error:
```
ablakley-r7, line 196: We should let errors bubble up to the run method.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+        :param alerts: list of alerts
+        :param state: state of the plugin
+        :return: list of deduped alerts, list of new hashes, and the highest timestamp
+        """
+
+        old_hashes, deduped_alerts, new_hashes, highest_timestamp = state.get(LAST_ALERT_HASH, []), [], [], ""
+        for index, alert in enumerate(alerts):
+
+            # todo - make method to convert
+            alert_time = alert.get(TIMESTAMP_KEY) / 1000
```
ablakley-r7, line 224: I think we should use a single timestamp value, if we are using created_at to query to filter and sort we should probably also use it to filter alerts for deduping.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+            return [], state, False, 500, PluginException(cause=error.cause, data=error)
+
+        except Exception as error:
+            print(f"{error = }")
+            return [], state, False, 500, PluginException(preset=PluginException.Preset.UNKNOWN, data=error)
+
+    ###########################
+    # Deduping
+    ###########################
+    def _dedupe_and_get_highest_time(self, alerts: list, start_time: str, state: dict) -> Tuple[list, list, str]:
```
ablakley-r7, line 210: We don't seem to call this method yet, will we have logic to get the old_hashes?

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+        if deduped_alerts:
+            # alert results are already sorted by CS API, so get the latest timestamp
+            highest_timestamp = deduped_alerts[-1].get(TIMESTAMP_KEY)
+
+            for alert in deduped_alerts:
+                if alert.get(TIMESTAMP_KEY) == highest_timestamp:
+                    new_hashes.append(hash_sha1(alert))
+                    self.logger.info(f"Hashed latest event with timestamp {highest_timestamp}.")
+
+            self.logger.debug(f"Highest timestamp is {highest_timestamp}")
```
ablakley-r7, line 248: We'll need to set this highest timestamp in state to use in future queries if we are no longer paginating but are catching up.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+        saved_time = state.get(LAST_ALERT_TIME)
+
+        log_msg = ""
+        if not saved_time:
+            log_msg += "No previous alert time within state. "
+            custom_timings = custom_config.get(LAST_ALERT_TIME, {})
+            custom_date = custom_timings.get("date")
+            custom_hours = custom_timings.get("hours", DEFAULT_LOOKBACK_HOURS)
+            start = datetime(**custom_date) if custom_date else (now - timedelta(hours=custom_hours))
+            state[LAST_ALERT_TIME] = start.strftime(TIME_FORMAT)
```
ablakley-r7, line 287: I don't think we need to stringify our dates, a datetime object should be readable in logs and state and also easy to translate into epoch time for our queries.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/util/api.py
```diff
@@ -182,10 +198,11 @@ def get_alerts(
         )
 
     def get_incidents(
-        self, from_time: int, to_time: int, time_sort_field: str = "creation_time", filters: List = None
+        self, from_time: int = None, to_time: int = None, time_sort_field: str = None, filters: List = None
```
ablakley-r7, line 201: Why have we stopped sorting incidents by creation_time?

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/util/api.py
```diff
@@ -182,10 +198,11 @@ def get_alerts(
...
     ) -> List[Dict]:
         endpoint = "/public_api/v1/incidents/get_incidents/"
         response_incidents_field = "incidents"
+        print(f"Get Incidents {from_time = }\n{to_time = }\n{time_sort_field = }\n{filters = }")
```
ablakley-r7, line 205: I think some of these can be reverted now we are using alerts?

plugins/palo_alto_cortex_xdr/plugin.spec.yaml
```diff
@@ -1038,62 +1038,13 @@ triggers:
...
-        required: false
+  monitor_alerts:
+    title: Monitor Alerts
+    description: Monitor alerts
     output:
-      events:
-        title: Incident Events
-        description: Incident events
-        type: "[]incident"
+      incidents:
```
ablakley-r7, line 1045:
```suggestion
      alerts:
```

ablakley-r7 (changes requested):

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
joneill-r7: Looks like the same for `parameters` defined on line 57 isn't used

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+
+        except PluginException as error:
+            self.logger.error(
+                f"A PluginException has occurred. Status code 500 returned. Error: {error.cause}. "
+                f"Existing state: {existing_state}"
+            )
+            return [], existing_state, False, 500, PluginException(cause=error.cause, data=error)
+
+        except Exception as error:
+            print(f"{error = }")
```
igorski-r7, line 101: Another one 😄

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+        except PluginException as error:
+            self.logger.error(
+                f"A PluginException has occurred. Status code 500 returned. Error: {error.cause}. "
+                f"Existing state: {state}"
+            )
+            return [], state, False, 500, PluginException(cause=error.cause, data=error)
+
+        except Exception as error:
+            print(f"{error = }")
+            return [], state, False, 500, PluginException(preset=PluginException.Preset.UNKNOWN, data=error)
```
igorski-r7, line 205: I think it's too complex. I suggest splitting it up into a couple smaller methods.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+            alert_time = alert.get(TIMESTAMP_KEY) / 1000
+            alert_time = datetime.fromtimestamp(alert_time, tz=timezone.utc).strftime(TIME_FORMAT)
+
+            if alert_time == start_time:
+                alert_hash = hash_sha1(alert)
+                if alert_hash not in old_hashes:
+                    deduped_alerts.append(alert)
+            elif alert_time > start_time:
+                deduped_alerts += alerts[index:]
+                print(f"{deduped_alerts= }")
```
igorski-r7, line 233: Print 😄

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+            comparison_date = datetime(**custom_lookback) if custom_lookback else default_date_lookback
+            comparison_date = comparison_date.replace(tzinfo=timezone.utc).strftime(TIME_FORMAT)
+            if comparison_date > saved_time:
+                self.logger.info(f"Saved time ({saved_time}) exceeds cut off, moving to ({comparison_date}).")
+                state[LAST_ALERT_TIME] = comparison_date
+                # pop state held time filters if they exist, incase customer has paused integration when paginating
+                state = self._drop_pagination_state(state)
+
+        start_time = state.get(LAST_ALERT_TIME)
+        print(f"{start_time = }")
```
igorski-r7, line 302: Another one 😄

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
cmcnally-r7: Debugging statements while we work on it

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_incident_events/task.py
rmurray-r7: > Not sure if this `monitor_incident_events` task is being used, should it be removed from the plugin?

It is not being used. We decided to create our task from scratch so I will delete it to resolve any confusion

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
rmurray-r7: I checked and it doesn't get set around lines 131/132. I checked the surrounding methods too 😕

ablakley-r7: It looks like the code has been moved around by additional updates.

Check these lines being duplicated.

        state[LAST_SEARCH_FROM] = search_from
        state[LAST_SEARCH_TO] = search_to

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,381 @@
...
+
+        custom_config = {}
+        state = {LAST_SEARCH_FROM: None, LAST_SEARCH_TO: None}
+
+        try:
+
+            now_time = self._get_current_time()
+            now = now_time - timedelta(minutes=15)
+            # last 15 minutes
+            end_time = now.strftime(TIME_FORMAT)
```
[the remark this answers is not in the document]
rmurray-r7: I have changed this now (as it was used prior for easy debugging) so the time gets set in the `custom_config`. As TimeDelta is used, I have to convert the UNIX to a DateTime object and then into a string (to add the millisecond field to the DateTime Object `[:-3] + "Z"`) and then back to DateTime for the TimeDelta to be carried out. Currently working on a smoother way to do this.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
rmurray-r7: Working on the exception handling at the moment. Will update when I made the changes 😄

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
rmurray-r7: Apologies, have corrected now!

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
rmurray-r7: I have made changes to use this method. Will look into whether the `old_hashes` are being used.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/util/api.py
rmurray-r7: I have made changes to now sort from `creation_time`

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
rmurray-r7: @cmcnally-r7 is finishing that up, once I make the final changes to the task code. Some of it has been refactored though! 😄

plugins/palo_alto_cortex_xdr/plugin.spec.yaml
```diff
@@ -1038,62 +1038,13 @@ triggers:
...
-        - "cron"
-        - "minutes"
-      default: minutes
-    state:
-      last_event_time:
-        title: Last Event Time
-        description: The datetime of the last retrieved incident event
-        type: string
\ No newline at end of file
+        example: '[ { } ]'
```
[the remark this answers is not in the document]
rmurray-r7: Done 😄

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
rmurray-r7: Looking into whether this method is needed or not.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,346 @@
...
+        """
+        Helper method to return the response JSON and total count
+        """
+        headers = self.connection.xdr_api.get_headers()
+
+        fqdn = self.connection.xdr_api.get_url()
+        url = f"{fqdn}{self.endpoint}"
+
+        request = requests.Request(method="post", url=url, headers=headers, json=post_body)
+        response = make_request(request, timeout=120)
```
ekelly-r7, line 326: You'll need to set the data_location here so that the response can be accessed in the exception 

i.e. 

```
        response = make_request(
            timeout=60,
            exception_data_location=ResponseExceptionData.RESPONSE,
        )
```

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
rmurray-r7: Update: It is now being used 😄

rmurray-r7:
Validator is failing due to the `connection_version` not being set to 4, even though it is.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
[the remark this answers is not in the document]
rmurray-r7: I tried and it does not work as it doesn't contain the `Z `for milliseconds. I have a managed to summarise the conversion into one line: 
```dt_now = datetime.strptime(
            datetime.fromtimestamp(now / 1000, tz=timezone.utc).strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3] + "Z",
            "%Y-%m-%dT%H:%M:%S.%fZ")```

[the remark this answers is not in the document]
Comment (rmurray-r7) [reply plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py]: You're right! I changed it to your reccomendation and changed the timezone to UTC and it works, thanks! 😄

File: plugins/palo_alto_cortex_xdr/bin/icon_palo_alto_cortex_xdr [hunk 33]
```diff
@@ -6,7 +6,7 @@ from sys import argv
 
 Name = "Palo Alto Cortex XDR"
 Vendor = "rapid7"
-Version = "4.0.0"
+Version = "5.0.0"
```
dlaverty-r7, line 9: As discussed, can we double check if a major version bump is actually required?

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,316 @@
...
+                f"Paginating alerts: Saving state with existing filters: "
+                f"search_from = {search_from} "
+                f"search_to = {search_to} "
+                f"current_count = {state.get(CURRENT_COUNT)} "
+                f"total_count = {total_count}"
+            )
+            state[LAST_SEARCH_TO] = search_to
+            state[LAST_QUERY_TIME] = start_time
+            state[LAST_SEARCH_FROM] = search_from
+            state[QUERY_START_TIME] = start_time
```
ekelly-r7, line 127: Is this `state[QUERY_START_TIME]` necessary if it's the same as `state[LAST_QUERY_TIME]`? both are set to `start_time`? 

The only usage I can see is on line 213/214 but it seems like we don't need one of them

cmcnally-r7: Will look into it 👍🏻

rmurray-r7: Just released a commit to address this 😄 
@ekelly-r7

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/util/api.py
```diff
@@ -419,3 +426,34 @@ def _get_endpoint_type(self, endpoint_info):
...
+        request = requests.Request(method="post", url=url, headers=headers, json=post_body)
+        response = make_request(_request=request, timeout=60, exception_data_location=ResponseExceptionData.RESPONSE)
+
+        response = extract_json(response)
+        total_count = response.get("reply", {}).get("total_count")
+        results_count = response.get("reply", {}).get("result_count", 0)
+        results = response.get("reply", {}).get("alerts", [])
+
+        # They physically return None so we can't default 0 in .get()
+        if total_count is None:
```
ablakley-r7, line 452: I think we can default this above instead

total_count = response.get("reply", {}).get("total_count", 0)

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,316 @@
...
+
+        results, results_count, total_count = self.connection.xdr_api.get_response_alerts(post_body)
+
+        state[CURRENT_COUNT] = state.get(CURRENT_COUNT, 0) + results_count
+
+        new_alerts, last_alert_hashes, last_alert_time = self._dedupe_and_get_highest_time(results, start_time, state)
+
+        is_paginating = state.get(CURRENT_COUNT) < total_count
+
+        has_more_pages = False
```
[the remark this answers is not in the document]
rmurray-r7: We moved the `has_more_pages` flag to only be set in the if or else statement (and not defaulted before)

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,316 @@
...
+        else:
+            first_run = False
+
+            # check if we have held the TS beyond our max lookback
+            lookback_days = custom_config.get(f"{LAST_ALERT_TIME}_days", MAX_LOOKBACK_DAYS)
+            default_date_lookback = dt_now - timedelta(days=lookback_days)  # if not passed from CPS create on the fly
+
+            custom_lookback = custom_config.get(f"max_{LAST_ALERT_TIME}", {})
+            comparison_date = datetime(**custom_lookback) if custom_lookback else default_date_lookback
+            comparison_date = comparison_date.replace(tzinfo=timezone.utc).strftime(TIME_FORMAT)
```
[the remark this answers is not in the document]
rmurray-r7: Yeah looking into a way we can create a DT object with milliseconds instead of having to convert it to a string to do so.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/util/api.py
cmcnally-r7: No we can't, ran into this issue yesterday and if it returns `None` then it doesn't fall back to `0`, causing issues.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,312 @@
...
+    ###########################
+    # Datetime Helpers
+    ###########################
+    def convert_unix_to_datetime(self, unix_time: int) -> datetime:
+        return datetime.fromtimestamp(unix_time / 1000, tz=timezone.utc)
+
+    def convert_to_unix(self, date_time: datetime) -> int:
+        return int(datetime.strptime(date_time, TIME_FORMAT).timestamp()) * 1000
+
+    def _drop_pagination_state(self, state: dict) -> Dict[str, Union[int, str, list]]:
```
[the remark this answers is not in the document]
cmcnally-r7: Do you mean the whole functionality? Or just to make the input `state: dict = {}`
We can't reset the entire state because we don't drop every key

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,312 @@
...
+            filters.append({"field": self.time_sort_field, "operator": "gte", "value": start_time})
+            filters.append({"field": self.time_sort_field, "operator": "lte", "value": end_time})
+
+        post_body = self.build_post_body(search_from=search_from, search_to=search_to, filters=filters)
+
+        results, results_count, total_count = self.connection.xdr_api.get_response_alerts(post_body)
+
+        state[CURRENT_COUNT] = state.get(CURRENT_COUNT, 0) + results_count
+
+        new_alerts, last_alert_hashes, last_alert_time = self._dedupe_and_get_highest_time(results, start_time, state)
```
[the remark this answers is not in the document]
cmcnally-r7: Whats the point in renaming this?

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,316 @@
...
+        timestamp.
+
+        :param alerts: list of alerts
+        :param state: state of the plugin
+        :return: list of deduped alerts, list of new hashes, and the highest timestamp
+        """
+
+        old_hashes, deduped_alerts, new_hashes, highest_timestamp = state.get(LAST_ALERT_HASH, []), [], [], ""
+        for index, alert in enumerate(alerts):
+            alert_time = alert.get(TIMESTAMP_KEY)
```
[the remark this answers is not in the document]
ablakley-r7: My bad!

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,312 @@
...
+        else:
+            state = self._drop_pagination_state(state)
+            self.logger.info(f"Remaining alerts: {len(new_alerts)}, is_paginating: {is_paginating}, \nstate: {state}")
+            has_more_pages = False
+
+        # add the last alert time to the state if it exists
+        # if not then set to the last queried time to move the filter forward
+        state[LAST_ALERT_TIME] = last_alert_time if last_alert_time else end_time
+        # update hashes in state only if we've got new ones
+        state[LAST_ALERT_HASH] = last_alert_hashes if last_alert_hashes else state.get(LAST_ALERT_HASH, [])
```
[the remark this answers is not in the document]
cmcnally-r7: When we get to the outer pagination, we can set this to empty list rather than `else grab the last alert hash`

plugins/palo_alto_cortex_xdr/unit_test/parameters/monitor_alerts.json.resp
```diff
@@ -0,0 +1,183 @@
...
+                    "action_file_macro_sha256": null,
+                    "action_file_md5": null,
+                    "action_file_name": null,
+                    "action_file_path": null,
+                    "action_file_sha256": null,
+                    "fw_device_name": null,
+                    "fw_rule_id": null,
+                    "fw_rule": null,
+                    "fw_serial_number": null,
+                    "agent_fqdn": "FL-ATR-DC3-04.attractions.disney.com",
```
ekelly-r7, line 54: Just double-checking that these alert samples (and throughout this PR) are not from real alerts? 

we should only be adding test/dummy data

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/util/api.py
cmcnally-r7: Throwing in result count blah blah for this too

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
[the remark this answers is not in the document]
cmcnally-r7: nvm, after discussing, renaming it

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/connection/connection.py
```diff
@@ -35,3 +40,46 @@ def clean_up_fqdn(self, fqdn):
...
+        start_time = datetime.fromisoformat(start_time)
+        # Timestamp in logs is in Miliseconds
+        start_time = int(start_time.timestamp() * 1000)
+
+        end_time = now.isoformat()
+        end_time = datetime.fromisoformat(end_time)
+        end_time = int(end_time.timestamp() * 1000)
+
+        try:
+            _ = self.xdr_api.get_alerts(start_time, end_time)
```
ablakley-r7, line 60: Just noticed this calls get_alerts whereas the task calls get_response_alerts. I think this and the error handling might need changed. Might also be worth checking the messages raised in the existing _post_to_api method need applied to the new make_request within get_response_alerts.

plugins/palo_alto_cortex_xdr/plugin.spec.yaml
```diff
@@ -4,13 +4,13 @@ products: [insightconnect]
 name: palo_alto_cortex_xdr
 title: Palo Alto Cortex XDR
 description: Stop modern attacks with the industry's first extended detection and response platform that spans your endpoints, network and cloud data
-version: 3.0.0
-connection_version: 2
+version: 4.1.0
```
ablakley-r7, line 7: Version 4.0.0 isn't released yet so good news is we can keep this at 4.0.0 and put the two updates out together.

cmcnally-r7: Really good spot Oo

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/connection/connection.py
cmcnally-r7: Thanks for noticing this

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,312 @@
...
+        return post_body
+
+    ###########################
+    # Datetime Helpers
+    ###########################
+    def convert_unix_to_datetime(self, unix_time: int) -> datetime:
+        return datetime.fromtimestamp(unix_time / 1000, tz=timezone.utc)
+
+    def convert_to_unix(self, date_time: datetime) -> int:
+        return int(datetime.strptime(date_time, TIME_FORMAT).timestamp()) * 1000
```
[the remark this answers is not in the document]
ablakley-r7: If we accept a datetime object

return int(datetime.timestamp() * 1000)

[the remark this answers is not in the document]
cmcnally-r7: I tried to implement this with no luck

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/util/api.py
```diff
@@ -419,3 +426,38 @@ def _get_endpoint_type(self, endpoint_info):
...
+            total_count = 0
+
+        # Also including results count just to be safe.
+        if results_count is None:
+            results_count = 0
+
+        return results, results_count, total_count
+
+
+class AuthenticationError(Exception):
```
ekelly-r7, line 462: Is this being used anywhere? could be removed if it's not

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -0,0 +1,312 @@
...
+        post_body = {
+            "request_data": {
+                "search_from": search_from,
+                "search_to": search_to,
+                "sort": {"field": self.time_sort_field, "keyword": "asc"},
+                "filters": filters,
+            }
+        }
+
+        self.logger.info(f"Post Body: {post_body}")
```
ekelly-r7, line 257: I don't think we need to log the Post Body as an `info`, better to remove or log as debug

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/util/api.py
cmcnally-r7: No it's not, not sure why this is even here, I'll remove it now

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
cmcnally-r7: Will change this to debug

ekelly-r7:
Looks good, @ablakley-r7 comments around plugin exceptions may need to be addressed before merging

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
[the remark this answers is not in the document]
ablakley-r7: I think in usage this method is being given a string, that should work if we take away the strftime input.

my_unix = 1726822543074
my_datetime = datetime.datetime.fromtimestamp(my_unix / 1000, tz=timezone.utc)
my_new_unix = int(my_datetime.timestamp() * 1000)
print(my_new_unix) #will print 1726822543074

ekelly-r7 (approved):

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/connection/connection.py
```diff
@@ -35,3 +40,56 @@ def clean_up_fqdn(self, fqdn):
...
+        }
+
+        try:
+            _, _, _ = self.xdr_api.get_response_alerts(post_body)
+            message = "The connection test to Palo Alto Cortex was successful"
+            self.logger.info(message)
+            return {"success": True}, message
+
+        except PluginException as error:
+            if error.cause in ("401", "402", "403", "404"):
```
ablakley-r7, line 76: I think this error handling will have to change, the get_response_alerts doesn't use the same response handler as the other API call methods.

cmcnally-r7:
> Looks good, @ablakley-r7 comments around plugin exceptions may need to be addressed before merging
> 
> Validator is also failing on connection_version, is that expected?

Yeah we need to raise a ticket around this, since the validator is looking for `connection_version + 1`. Since `v4.0.0` has not be released, and we're releasing this alongside it, we're keeping the connection version the same.

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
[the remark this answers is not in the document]
cmcnally-r7: Finally got it, great spot!

ablakley-r7 (approved):

igorski-r7 (approved):

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/connection/connection.py
```diff
@@ -35,3 +40,55 @@ def clean_up_fqdn(self, fqdn):
...
+            message = "The connection test to Palo Alto Cortex was successful"
+            self.logger.info(message)
+            return {"success": True}, message
+
+        except PluginException as error:
+            return_message = f"The connection test to Palo Alto Cortex failed.\n"
+            return_message += f"This failure was caused by: '{error.cause}'\n"
+            return_message += f"Error assistance: {error.assistance}\n"
+
+            self.logger.info(f"error = {error.cause}")
```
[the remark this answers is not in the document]
cmcnally-r7: Good spot, forgot to remove it

PR #2806, opened 2024-09-20
[SOAR-17746] - Fix for get_url | Refresh plugin

ekelly-r7:
### Description

  - Fix for issue where a PluginException occurs when running get_file action multiple times for the same URL
  - https://rapid7.atlassian.net/browse/SOAR-17746

When URL contents are downloaded, a meta file that is saved in the cache which contains a `last-modified` and `etag` value. This value can be `null`, e.g. 

```
➜  cache cat 7c774392df25c0990a337dc8b862b783cd881b51.meta
{"url": "https://www.bing.com/", "last-modified": null, "etag": null, "file": "7c774392df25c0990a337dc8b862b783cd881b51.file"}%
```

Fix to explicitly set `etag` and `last-modified` None header values to `""` so that it will work with `request.urlopen` (which is expecting a 'str' or 'bytes' value)


plugins/get_url/komand_get_url/util/utils.py
```diff
@@ -79,13 +79,18 @@ def check_prefix_and_download(
         if insightconnect_plugin_runtime.helper.check_cachefile(meta.get("metafile")):
             headers = self.check_url_meta_file(meta)
 
+        # Caching headers from meta info can be None
+        # request.urlopen expects str or bytes
+        etag = headers.get("etag", "") or ""
+        last_modified = headers.get("last-modified", "") or ""
+
```
ekelly-r7, line 86: Fix here to explicitly set these header values to `""` when they are `null` in the meta cache file

Otherwise `requests.urlopen` returns an exception

joneill-r7 (approved):

igorski-r7 (approved):

ablakley-r7 (approved):

PR #2810, opened 2024-09-20
[SOAR-17226] IPSTACK Connection Test added & SDK Bump/ Plugin Refresh

rmurray-r7:
### Description

The connection test was always passing as it was hardset to  `return {"success": True}`. As there was no `api.py` to test the connection through, tested through opening the URL as the SecretKey is used in the URL (and will fail is SecretKey is invalid).

As the user does not see the Exception data, I hardset the values.

Ticket: https://rapid7.atlassian.net/browse/SOAR-17726

  - SDK Bump to 6.1.2 & Plugin Refresh
  - Connection Test added

### Testing

Tested by passing a valid and invalid SecretKey:

- Invalid
image

- Valid
image


plugins/ipstack/icon_ipstack/connection/connection.py
```diff
@@ -13,4 +16,18 @@ def connect(self, params):
         self.token = params.get(Input.CRED_TOKEN).get("secretKey")
```
igorski-r7, line 16:
```suggestion
        self.token = params.get(Input.CRED_TOKEN, {}).get("secretKey", "").strip()
```

plugins/ipstack/icon_ipstack/connection/connection.py
```diff
@@ -13,4 +16,18 @@ def connect(self, params):
...
 
     def test(self):
-        return {"success": True}
+        # Sending a request using the secretKey to test if it is valid
+        url = "http://api.ipstack.com/" + "rapid7.com" + "?access_key=" + self.token + "&output=json"
+
+        invalid_status_codes = ["404", "101", "102", "104", "106"]
```
rbowden-r7, line 22: Might be worth adding a comment here to say that these 1xx are custom status code returned from ipstack as in the future people might question why we are checking against these

rmurray-r7: Good idea, will add now!

plugins/ipstack/icon_ipstack/connection/connection.py
```diff
@@ -13,4 +16,18 @@ def connect(self, params):
...
+        resp = insightconnect_plugin_runtime.helper.open_url(url)
+        dic = json.loads(resp.read())
+        if any(invalid_status_codes) in dic or "error" in dic:
+            raise PluginException(
+                data=dic,
+                assistance="Please check your credentials and try again. If the issue persists, please contact support.",
+                cause="An error has occurred. Please check your credentials are correct and that you have not exceeded the lookup limits.",
+            )
+        else:
+            return {"success": True}
```
igorski-r7: I suggest using `example.com` instead of `rapid7.com`.

```suggestion
        url = f"http://api.ipstack.com/example.com?access_key={self.token}&output=json"
        response = insightconnect_plugin_runtime.helper.open_url(url)
        parsed_response = json.loads(response.read())
        invalid_status_codes = ["404", "101", "102", "104", "106"]
        if any(invalid_status_codes) in parsed_response or "error" in dic:
            raise PluginException(
                data=parsed_response,
                assistance="Please check your credentials and try again. If the issue persists, please contact support.",
                cause="An error has occurred. Please check your credentials are correct and that you have not exceeded the lookup limits.",
            )
        return {"success": True}
```

plugins/ipstack/icon_ipstack/connection/connection.py
```diff
@@ -13,4 +16,19 @@ def connect(self, params):
         self.token = params.get(Input.CRED_TOKEN).get("secretKey")
 
     def test(self):
-        return {"success": True}
+        # Sending a request using the secretKey to test if it is valid
+        url = "http://api.ipstack.com/" + "rapid7.com" + "?access_key=" + self.token + "&output=json"
```
igorski-r7, line 20: I suggest using f-strings here

plugins/ipstack/icon_ipstack/connection/connection.py
```diff
@@ -13,4 +16,19 @@ def connect(self, params):
...
+
+        # These contain custom status codes (101, 102, 104, 106) returned from IPStack, as seen in the lookup action.py
+        invalid_status_codes = ["404", "101", "102", "104", "106"]
+
+        resp = insightconnect_plugin_runtime.helper.open_url(url)
```
igorski-r7, line 25:
```suggestion
        response = insightconnect_plugin_runtime.helper.open_url(url)
```

plugins/ipstack/icon_ipstack/connection/connection.py
```diff
@@ -13,4 +16,19 @@ def connect(self, params):
...
+        dic = json.loads(resp.read())
```
igorski-r7, line 26:
```suggestion
        parsed_response = json.loads(response.read())
```

plugins/ipstack/icon_ipstack/connection/connection.py
```diff
@@ -13,4 +16,19 @@ def connect(self, params):
...
+        if any(invalid_status_codes) in dic or "error" in dic:
```
igorski-r7, line 27: Is that condition correct? I mean the `any()` built-in returns `True` if any condition specified in iterable is `True`. As it looks like `if True/False in dic` when executed 😄

plugins/ipstack/icon_ipstack/connection/connection.py
```diff
@@ -13,4 +16,19 @@ def connect(self, params):
...
+            raise PluginException(
+                data=dic,
```
igorski-r7, line 29:
```suggestion
                data=parsed_response,
```

plugins/ipstack/icon_ipstack/connection/connection.py
```diff
@@ -13,4 +16,19 @@ def connect(self, params):
...
+                assistance="Please check your credentials and try again. If the issue persists, please contact support.",
+                cause="An error has occurred. Please check your credentials are correct and that you have not exceeded the lookup limits.",
+            )
+        else:
+            return {"success": True}
```
igorski-r7, line 34: Not necessary `else`
```suggestion
        return {"success": True}
```

plugins/ipstack/icon_ipstack/connection/connection.py
```diff
@@ -13,4 +16,19 @@ def connect(self, params):
         self.token = params.get(Input.CRED_TOKEN).get("secretKey")
 
     def test(self):
-        return {"success": True}
+        # Sending a request using the secretKey to test if it is valid
+        url = "http://api.ipstack.com/" + "rapid7.com" + "?access_key=" + self.token + "&output=json"
+
+        # These contain custom status codes (101, 102, 104, 106) returned from IPStack, as seen in the lookup action.py
+        invalid_status_codes = ["404", "101", "102", "104", "106"]
```
rmurray-r7: The reason we have to use status codes in this way is due to IPStack always returning a 200, even when failing. This is the output for a failed test :

"rapid7/IPStack:3.0.2\nresponse.msg= 'OK'\nresponse.getcode()= 200\nparsed_response= {'success': False, 'error': {'code': 101, 'type': 'missing_access_key', 'info': 'You have not supplied an API Access Key. [Required format: access_key=YOUR_ACCESS_KEY]'}}\n".

I have changed it to check the `success` key as it differs when the test is valid or not.

ekelly-r7 (approved):

igorski-r7 (approved):

PR #2955, opened 2024-11-08
[Palo Alto] Error 401 correctly

nmccullagh-r7:
## Proposed Changes
https://rapid7.atlassian.net/browse/SOAR-18176


plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
```diff
@@ -61,6 +61,15 @@ def run(self, params={}, state={}, custom_config: dict = {}):  # pylint: disable
 
             return response, state, has_more_pages, 200, None
 
+        except APIException as error:
+            return (
+                [],
+                {},
```
ekelly-r7, line 67: Do we want to return an empty dict for state here? or should just return `existing_state` ?

plugins/palo_alto_cortex_xdr/unit_test/test_get_file_quarantine_status.py
```diff
@@ -0,0 +1,19 @@
...
+import logging
+
+
+class TestGetFileQuarantineStatus(TestCase):
+    def test_get_file_quarantine_status(self):
+        """
+        DO NOT USE PRODUCTION/SENSITIVE DATA FOR UNIT TESTS
+
+        TODO: Implement test cases here
+        """
```
ekelly-r7, line 19: prob worth removing this test file as it's just a stub?

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/tasks/monitor_alerts/task.py
nmccullagh-r7: existing state isn't used for anything in an error case in pif 🤷

ekelly-r7: yeah fair enough, just a np more than anything as that's the (albeit redundant) pattern that we usually follow

ekelly-r7 (approved):

plugins/palo_alto_cortex_xdr/icon_palo_alto_cortex_xdr/util/api.py
```diff
@@ -473,12 +474,21 @@ def _handle_401(self, response: Response, url: str, post_body: dict):
...
-        if response.status_code == 401:
-            self.logger.info(f"Received HTTP {response.status_code}, re-authenticating to Palo Alto Cortex XDR")
+        if response.status_code in [401, 403]:
+            self.logger.info(f"Received HTTP {response.status_code}, attempting to refresh token")
 
             new_headers = self._advanced_authentication(api_key=self.api_key, api_key_id=self.api_key_id)
             new_response = self.build_request(url=url, headers=new_headers, post_body=post_body)
 
+            if new_response.status_code in [401, 403]:
+                self.logger.info("Token is still invalid, raising exception")
```
ekelly-r7, line 484: np: I think should be `self.logger.error`

PR #2974, opened 2024-11-21
[SOAR-18258] Whois - Update address action

rmurray-r7:
### Description

A customer was experiencing an issue when proving an RIPE IP, the `orgname` would not return when using ICON, but it would return when using WhoIs website. This is cause the orgname pointed to the description.

As the description did not have an existing field, as the orgname pointed to it, I added a description field too.

[SI TICKET](https://rapid7.atlassian.net/browse/SI-26530)

  - Update Address action (add `description` output field. Correct mapping for `orgname` field)
  - SDK bump to 6.2.0

## PR Requirements

### Testing

When using the same IP the user used, it now returns the `orgname`:
image

As the above IP does not have a description, when using another RIPE IP, it returns the description too:
image

When specifiying the registrar as RIPE, the mapping also works:
image

### DIFF
--- a/plugins/whois/.CHECKSUM
+++ b/plugins/whois/.CHECKSUM
@@ -1,7 +1,7 @@
 {
-	"spec": "651e527db096e0dd0ac0e9896a4b2756",
-	"manifest": "e4ca9901883c704390367efe46454f05",
-	"setup": "4f04c80ff4c28070c7ec483f1ea2b876",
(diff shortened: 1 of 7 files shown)


rmurray-r7:
Validator is expected to fail, due to the **-y** flag in the Dockerfile

rbowden-r7 (approved):

igorski-r7 (approved):

PR #2976, opened 2024-11-22
[SOAR-18265] Palo Alto Pan OS - Fix add_address_object_to_group action

rmurray-r7:
### Description

[TICKET](https://rapid7.atlassian.net/browse/SOAR-18265)

Investigating an [SI issue](https://rapid7.atlassian.net/browse/SI-26543), when debugging it seems the `.join() `operator should not be used due to it adding significantly large amount of duplicate data which leads to the following error:

`insightconnect_plugin_runtime.exceptions.LoggedException: ('Connection aborted.', BadStatusLine('<html>\r\n'))
`

The current plugin, reads the existing IP's in a list then appends the new one onto it, but the `.join()` operator does this:

image

This is due to joining all of the previous IP's along with the new one in each iteration (eg if the list has abc and next element is def the input will be abcdef and not def)

Changing the `.join()` to just add the string via `+=` resolves this:

image

**This does not resolve the [SI issue](https://rapid7.atlassian.net/browse/SI-26543). That is caused by Palo Alto and not us :)** 

  - Fix `add_address_object_to_group` action (duplicate appending)
  - SDK bump to 6.2.0
  - Bumping requirements.txt

### Testing

Testing it locally it now passes:
image

#### Unit Tests
All unit tests pass:
image

#### Snyk Test
No vulns:
image

### DIFF
--- a/plugins/palo_alto_pan_os/requirements.txt
+++ b/plugins/palo_alto_pan_os/requirements.txt
@@ -7,5 +7,4 @@ dicttoxml==1.7.4
 validators==0.34.0
 IPy==1.01
 parameterized==0.9.0
-jsonschema==4.21.1
 setuptools==70.0.0
--- a/plugins/palo_alto_pan_os/requirements.txt
+++ b/plugins/palo_alto_pan_os/requirements.txt
@@ -4,8 +4,8 @@
 gunicorn==22.0.0
 xmltodict==0.12.0
 dicttoxml==1.7.4
-validators==0.22.0
+validators==0.34.0
 IPy==1.01
-parameterized==0.8.1
-jsonschema==3.2.0
+parameterized==0.9.0
+jsonschema==4.21.1
 setuptools==70.0.0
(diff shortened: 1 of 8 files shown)


plugins/palo_alto_pan_os/requirements.txt
```diff
@@ -4,8 +4,8 @@
 gunicorn==22.0.0
 xmltodict==0.12.0
 dicttoxml==1.7.4
-validators==0.22.0
+validators==0.34.0
 IPy==1.01
-parameterized==0.8.1
-jsonschema==3.2.0
+parameterized==0.9.0
+jsonschema==4.21.1
```
igorski-r7, line 10: Isn't that included to the SDK. If so, I think we can remove it from here.

rmurray-r7: I am not too sure. Is it? I know `requests` is but not too sure about `jsonschema`?

igorski-r7 (approved):

rbowden-r7 (approved):

PR #3035, opened 2025-01-08
Okta plugin.spec sync

rmurray-r7:
### Description

Updating the `plugin.spec` to be in sync with the `help.md`. There is no way to add the extra details in the plugin.spec so the user knows the action output when there is no users/groups, but I think it can be removed as it's common sense 😅 

This will get bundled alongside #3012 

  - Updating `plugin.spec`

### DIFF
--- a/plugins/okta/komand_okta/tasks/monitor_logs/schema.py
+++ b/plugins/okta/komand_okta/tasks/monitor_logs/schema.py
@@ -43,9 +43,7 @@ class MonitorLogsOutput(insightconnect_plugin_runtime.Output):
   "type": "array",
   "title": "Logs",
   "description": "All system logs within the specified time range",
-  "items": {
-    "type": "object"
-  },
+  "items": {},
   "required": [
     "logs"
   ],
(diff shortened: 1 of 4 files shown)


plugins/okta/komand_okta/tasks/monitor_logs/schema.py
```diff
@@ -43,9 +43,7 @@ class MonitorLogsOutput(insightconnect_plugin_runtime.Output):
   "type": "array",
   "title": "Logs",
   "description": "All system logs within the specified time range",
-  "items": {
-    "type": "object"
-  },
+  "items": {},
```
igorski-r7, line 46: Wondering if that should be `{}`? I remember it was failing on schema time ago 😄

rmurray-r7: In the tooling I have updated it so all task outputs are like this. It was failing when it was `$ref : {}` but @ekelly-r7 found having it set to  `"items": {}` works as it allows for different types of outputs to be in the task 😄

igorski-r7: Sounds good! 😄 Thanks!

igorski-r7 (approved):

ablakley-r7 (approved):

PR #3044, opened 2025-01-13
[SOAR-18526] String (fedRAMP) plugin.spec sync

rmurray-r7:
### Description

Updating the `plugin.spec `to be inline with the` help.md`. There is a small amount of data in the existing `help.md` which cannot be put in the spec file to carry over but I don't think it is vital information (other input examples).
  
- Updating `plugin.spec` to be in sync with `help.md `

This will get merged into this PR:
- #3006 

Troubleshooting output:
image


plugins/string/help.md
```diff
@@ -206,16 +206,8 @@ Example output:
...
-
-Any input requiring more than a single key:value pair, e.g. `USER=Bob` needs to use the `block_delimiter` option.
-In this case, the input string is split by the `block_delimiter` character first, and the resulting items are then split
-by the `string_delimiter` option. Stripping of double-quotes is automatically applied in this situation for each item before the plugin returns it.
-
-The [output schema object](https://docs.komand.com/v0.42.1/docs/python-script-plugins#section-configure-the-plugin-output-schema) on the action's page can be modified to pre-populate the workflow with the names of the keys.
-It allows users the ability to use the green selector and choose a specific variable later in the workflow by name.
-[Input templating](https://docs.komand.com/docs/input-templating) would need to be used to obtain variables by name otherwise.
+  
+This action is used to convert a string to an object containing key:value strings
```
joneill-r7, line 210: chatted on slack around this - can we try to include the example/description around the input still here but the more in depth example in the troubleshooting just before the current troubleshooting point?

joneill-r7 (approved):

igorski-r7 (approved):

PR #3137, opened 2025-02-14
[SOAR-17182] Mimecast

ablakley-r7:
### Description

  - Update to Mimecast API 2.0
  - Remove Task
  - Update SDK

## PR Requirements

#### Unit Tests

#### In-Product Tests

### Style

- [ ] For size, use the [slim SDK images](https://docs.rapid7.com/insightconnect/sdk-guide/#sdk-guide) when possible: ``rapid7/insightconnect-python-3-38-slim-plugin:{sdk-version-num}`` and ``rapid7/insightconnect-python-3-38-plugin:{sdk-version-num}``
- [ ] For docs, validate markdown with ``insight-plugin validate`` which calls ``icon_validate`` to lint ``help.md``

### Functional Checklist
- [ ] Work fully completed
- [ ] Functional
  - [ ] Any new actions/triggers include JSON [test files](https://docs.rapid7.com/insightconnect/style-guide/#tests) in the `tests/` directory created with `insight-plugin samples`
    * Single action/trigger `insight-plugin run -T tests/example.json --debug --jq`
    * All actions/triggers shortcut `insight-plugin run -T all --debug --jq` (use PR format at end)
    * Single action/trigger `insight-plugin run -R tests/example.json --debug --jq`
    * All actions/triggers shortcut `insight-plugin run --debug --jq` (use PR format at end)

### Assessment

1. Run `insight-plugin validate` and make sure everything passes
2. Run the assessment tool: `insight-plugin run -A`. For single action validation: `insight-plugin run tests/{file}.json -A`
3. Copy (`insight-plugin ... | pbcopy`) and paste the output in **a new post** on this PR


plugins/mimecast/komand_mimecast/connection/connection.py
```diff
@@ -38,28 +35,3 @@ def test(self):
                 data=response.get(FAIL_FIELD),
             )
         return {"success": True}
-
-    def test_task(self):
```
ekelly-r7, line 42: I assume when this goes GA, the Mimecast v1 tile will already be removed from Event Sources UI / products.json? 

What happens if someone who already has Mimecast v1 tries to run a connection test when this logic is removed?

igorski-r7 (approved):

plugins/mimecast/komand_mimecast/connection/connection.py
ablakley-r7: Yep we'll be waiting until V2 is live to make the switchover. My understanding is that these customers will be version pinned to the previous version but I'll double check.

plugins/mimecast/.CHECKSUM
```diff
@@ -57,11 +57,7 @@
 		},
 		{
 			"identifier": "connection/schema.py",
-			"hash": "5f2643a1df641cfac4a907361dd63556"
-		},
-		{
```
joneill-r7, line 62: I think we'll just need to be careful that this doesn't go out and remove the task if we have anything running in EA

plugins/mimecast/plugin.spec.yaml
```diff
@@ -891,47 +890,17 @@ types:
...
-    required: true
-  secret_key:
-    title: Application Secret Key
-    description: The application secret key
+  client_id:
+    title: Client ID
+    description: The Mimecast 2.0 Application Client ID
     type: credential_secret_key
-    example: FgHrtydiP4TynI+rTZF42Qu0FtGuhJtuNM5bDh82goJQHed9kJZ5t/ORwGnI5r2hkl/bzCosZ+KVapJFeaf3Yw==
+    example: ZA7vkbu7NqcfBcGrXyWW8Rzk2sv2un2DCY7GGCX4BFWgJBZM
```
joneill-r7, line 897: do we need this to be a secret key if the application ID wasn't? I think there's UI changes needed when we do this

plugins/mimecast/komand_mimecast/util/api.py
```diff
@@ -369,17 +212,26 @@ def _handle_rest_call(  # noqa: C901
...
-
-        payload = {DATA_FIELD: ([data] if data is not None else [])}
-
+        payload = data
         if meta_data is not None:
             payload[META_FIELD] = meta_data
-
+        if not authenticating:
+            payload = str({DATA_FIELD: ([data] if data is not None else [])})
+        if not existing_request:
```
joneill-r7, line 226: this looks like it's for the case of the token expired. are we not better removing the access_token at the end of any action? I think we've done this in other plugins so that the access token isn't lying around. then this flag could and above could use `if self.access_token ... else ...`

plugins/mimecast/komand_mimecast/util/api.py
```diff
@@ -396,9 +248,23 @@ def _handle_rest_call(  # noqa: C901
...
         if response.get(FAIL_FIELD):
-            self._handle_error_response(response)
+            if response.get(FAIL_FIELD, [{}])[0].get("code", "") == "token_expired":
+                self.logger.info("Token has expired, attempting re-authentication...")
+                self.access_token = None
+                self.authenticate()
+                request = response.request
+                return self._handle_rest_call(
+                    method=request.method,
+                    uri=request.url,
```
joneill-r7, line 258: or if you don't want the `else` statement above (comment around `if not existing_request`), we still have access to `URI` value in this scope?

plugins/mimecast/komand_mimecast/util/constants.py
```diff
@@ -1,19 +1,4 @@
...
-REGIONS = ["EU", "DE", "US", "CA", "ZA", "AU", "Offshore", "Sandbox", "USB", "USBCOM"]
-HOSTS = [
-    "eu-api",
-    "de-api",
-    "us-api",
-    "ca-api",
-    "za-api",
-    "au-api",
-    "je-api",
-    "sandbox-api",
```
joneill-r7, line 11: wondering if there's still a sandbox API?

joneill-r7 (commented):
looks good, just a few small comments and questions

plugins/mimecast/komand_mimecast/util/api.py
ablakley-r7: This is just when we call the authentication endpoint, we don't want to modify the data in the request in that instance.

plugins/mimecast/komand_mimecast/util/api.py
ablakley-r7: URI should be passed into the method in all calls so should be available.

plugins/mimecast/.CHECKSUM
ablakley-r7: Have confirmed that existing event sources will be version pinned based on the connection version.

plugins/mimecast/komand_mimecast/util/constants.py
ablakley-r7: Nope! No longer exists.

plugins/mimecast/.CHECKSUM
joneill-r7: makes sense! I was even talking about how this would be the case to someone the other day, so apparently forgot that when reviewing this 🤦

plugins/mimecast/plugin.spec.yaml
joneill-r7: Hadn't realised this was from a random slack thread that you weren't involved in. It was raised for the event source that we needed to make a change in the 'add credential' screen to override the default - [UI-28115](https://rapid7.atlassian.net/browse/UI-28115). Whereas if this is in a string type we won't need to? Unless we do think a client ID is a sensitive value?

plugins/mimecast/komand_mimecast/util/api.py
joneill-r7: Yeah I'm following the conditions and they work ok! Just wondering if it could simplified slightly so we aren't passing in extra params when we can do without them, but not a sticking point. My thinking is we could remove the following:
- `authenticating` this could be determined with `if not self.access_token` and do the same thing
- `existing_request` we could either pass in the `uri` and not `request.url` when the token has expired or for each call of `_handle_rest_call` we pass the full URL and not just the endpoint?

ablakley-r7: Sounds fair. self.access_token is destroyed before attempting to re-authenticate anyway here.

joneill-r7 (approved):
lgtm

PR #3167, opened 2025-02-21
[SOAR-18956] Mimecast V2 - Update hash limits

ablakley-r7:
### Description

  - Update hash limits for state to reduce state size
  - Update log limit to 7500 for all logs
  - Context: [SOAR-18956](https://rapid7.atlassian.net/browse/SOAR-18956)

## PR Requirements

#### Unit Tests

#### In-Product Tests

### Style

- [ ] For size, use the [slim SDK images](https://docs.rapid7.com/insightconnect/sdk-guide/#sdk-guide) when possible: ``rapid7/insightconnect-python-3-38-slim-plugin:{sdk-version-num}`` and ``rapid7/insightconnect-python-3-38-plugin:{sdk-version-num}``
- [ ] For docs, validate markdown with ``insight-plugin validate`` which calls ``icon_validate`` to lint ``help.md``

### Functional Checklist
- [ ] Work fully completed
- [ ] Functional
  - [ ] Any new actions/triggers include JSON [test files](https://docs.rapid7.com/insightconnect/style-guide/#tests) in the `tests/` directory created with `insight-plugin samples`
    * Single action/trigger `insight-plugin run -T tests/example.json --debug --jq`
    * All actions/triggers shortcut `insight-plugin run -T all --debug --jq` (use PR format at end)
    * Single action/trigger `insight-plugin run -R tests/example.json --debug --jq`
    * All actions/triggers shortcut `insight-plugin run --debug --jq` (use PR format at end)

### Assessment

1. Run `insight-plugin validate` and make sure everything passes
2. Run the assessment tool: `insight-plugin run -A`. For single action validation: `insight-plugin run tests/{file}.json -A`
3. Copy (`insight-plugin ... | pbcopy`) and paste the output in **a new post** on this PR

[SOAR-18956]: https://rapid7.atlassian.net/browse/SOAR-18956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/mimecast_v2/icon_mimecast_v2/tasks/monitor_siem_logs/task.py
```diff
@@ -205,6 +206,33 @@ def get_all_logs(
...
+        :type list:
+        :param new_logs: New values to hash and compare to existing list of hashes.
+        :type list:
+        :return: Hex digest of hash.
+        :rtype: Tuple[list, list]
+        """
+        new_logs_hashes = []
+        logs_to_return = []
+        # Limit the amount of log hashes saved in order to reduce state size
+        log_hash_save_start = len(new_logs) - LOG_HASH_SIZE_LIMIT
```
joneill-r7, line 223: Is this condition right? if we have less than 1000 logs in a response, it means that we won't be hashing anything to compare on the next run as it'll go into a minus value. We need a few more checks that `new_logs` is more than LOG_HASH_SIZE_LIMIT

ablakley-r7: So this means combined with lines below that when hashing each log, whenever we hit an index greater than the log_hash_save_start, we'll add it to the new hashes.

I had a max condition on this (0 or length -1000) but it's not useful as we will add the hash to the list regardless of the limit being 0 or below 0.

igorski-r7 (approved):

plugins/mimecast_v2/icon_mimecast_v2/tasks/monitor_siem_logs/task.py
ablakley-r7: Yes I thought the same, the only thing halting me on implementing it was the log type receipt is far and away the highest volume log type. 

My thinking is that if we're reducing the effectiveness of deduplication, which might only dedupe a very small number of logs in the first place, then is it worth splitting each log type out to check the hashes if two of the three log types only make up a very small portion of all events.

For instance in some testing, we can see over 40,000 receipt logs where we only get 3 for the others combined. I wasn't sure it was worth implementing logic to dedupe 3 logs over keeping the entire log set together and grabbing the last 1000 max.

Reasons why each event type is generated are below:

Receipt - where the MTA receives a new connection for an email, either from your organization's email infrastructure or the outside world.

For URL Protect, a log line is written each time a user clicks a link that has been rewritten by Mimecast in an email and has been found to be malicious.

For Attachment Protect, a log line is written for each file processed by the sandbox and found to be malicious.

plugins/mimecast_v2/icon_mimecast_v2/tasks/monitor_siem_logs/task.py
```diff
@@ -52,7 +53,7 @@ def run(self, params={}, state={}, custom_config={}):  # pylint: disable=unused-
             query_config = self.prepare_query_params(state.get(QUERY_CONFIG, {}), max_run_lookback_date, now_date)
             logs, query_config = self.get_all_logs(run_condition, query_config, page_size, thead_count)
             self.logger.info(f"TASK: Total logs collected this run {len(logs)}")
-            logs, log_hashes = compare_and_dedupe_hashes(state.get(LOG_HASHES, []), logs)
+            logs, log_hashes = self.compare_and_dedupe_hashes(state.get(LOG_HASHES, []), logs)
```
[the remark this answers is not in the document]
ablakley-r7: Testing and page limit is setting correctly for the log types. Start at 1 and moving up I can see the amount of logs returned per run increasing. Page limit refers to how many batches can be returned in one request, with each batch containing the actual logs.

We'll pass page_size to get_siem_batches which returns a list of pre-signed URLs up to that number that contain all our logs in JSON files.

plugins/mimecast_v2/icon_mimecast_v2/tasks/monitor_siem_logs/task.py
joneill-r7: I do think it would be worth implementing separating out the log hashes against their own log type, otherwise we're comparing hashes against a list (up to 3k for each hash type) when we know only 1k have the potential to match?

Thanks for the event type description also! Probably is unlikely to be receive high volume of the others, but would also mean we could be comparing the 1 or 2 logs against the list of 1k for a different log type (receipt).

That would mean in this example you mentioned: 

> "we can see over 40,000 receipt logs where we only get 3 for the others combined. I wasn't sure it was worth implementing logic to dedupe 3 logs over keeping the entire log set together and grabbing the last 1000 max."

The logic in this would check the next 40,000 (think this needs looked at anyway, but that's the next comment thread) receipt logs against the last 1k stored, and the 3 others against their corresponding log type and actually remove an accurate dupes?

plugins/mimecast_v2/icon_mimecast_v2/tasks/monitor_siem_logs/task.py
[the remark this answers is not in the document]
joneill-r7: I was having a similar discussion around another plugin and think if we've downloaded the file we might be better keeping the content but then stopping to process anymore if we've hit the limit. From memory on other talks we can exceed the 7500 being returned but this puts more stress on the CE which is managing multiple plugins. So if we could do some sort of optimisation on how much being returned it might better, although agreed it does then complicate mid-parsing the list of pre-signed URLs

plugins/mimecast_v2/icon_mimecast_v2/tasks/monitor_siem_logs/task.py
ablakley-r7: We had a chat to sum up. Split hashes per log type with a reduced count for low traffic log types.

plugins/mimecast_v2/icon_mimecast_v2/tasks/monitor_siem_logs/task.py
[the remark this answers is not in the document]
ablakley-r7: Updated. We log out the URL - it's better for debugging but keen to hear thoughts on whether to keep it.

igorski-r7 (approved):

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
```diff
@@ -79,15 +109,41 @@ def get_siem_batches(
...
+    def get_siem_logs_from_batch(self, url_and_position: Tuple[str, int]) -> Tuple[List[Dict], str]:
+        url, line_start = url_and_position
         response = requests.request(method=GET, url=url, stream=False)
         with gzip.GzipFile(fileobj=BytesIO(response.content), mode="rb") as file_:
             logs = []
             # Iterate over lines in the decompressed file, decode and load the JSON
-            for line in file_:
+            for _, line in enumerate(file_, start=line_start):
                 decoded_line = line.decode("utf-8").strip()
                 logs.append(json.loads(decoded_line))
```
joneill-r7, line 140: i'm just remembering we had issues on v1 that some files could contain malformed JSON and then we got stuck in a loop - should we allow for this in here again and continue to the next file?

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
```diff
@@ -40,21 +41,50 @@ def authenticate(self) -> None:
...
+        logs = manager.list()
+        lock = manager.Lock()
         with Pool(max_threads) as pool:
-            batch_logs = pool.imap(self.get_siem_logs_from_batch, batch_download_urls)
-            for result in batch_logs:
-                if isinstance(result, (List, Dict)):
-                    logs.extend(result)
+            result = pool.imap(self.get_siem_logs_from_batch, pool_data)
+            for batch_logs, url in result:
+                if isinstance(batch_logs, (List, Dict)):
```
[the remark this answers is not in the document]
ablakley-r7: Single value files can come in with a single JSON object and no list.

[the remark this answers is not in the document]
ablakley-r7: Actually looking again, great point. This is a hold over from previous code that handled the combining of the list / dict.

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
```diff
@@ -79,15 +109,41 @@ def get_siem_batches(
...
+        return pool_data
+
+    def get_siem_logs_from_batch(self, url_and_position: Tuple[str, int]) -> Tuple[List[Dict], str]:
+        url, line_start = url_and_position
         response = requests.request(method=GET, url=url, stream=False)
         with gzip.GzipFile(fileobj=BytesIO(response.content), mode="rb") as file_:
             logs = []
             # Iterate over lines in the decompressed file, decode and load the JSON
-            for line in file_:
+            for _, line in enumerate(file_, start=line_start):
```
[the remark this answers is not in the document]
ablakley-r7: So these would be the same thing in this case, the counter is the line we want to start with (0, or a saved index for saved files). Each line is a JSON object.

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
```diff
@@ -79,15 +109,41 @@ def get_siem_batches(
...
+            list_of_batches = list_of_batches[
+                next((index for index, url in enumerate(list_of_batches) if saved_url in url), len(list_of_batches)) :
+            ]
+
+        pool_data = [(url, starting_position if saved_url and saved_url in url else 1) for url in list_of_batches]
+        return pool_data
+
+    def get_siem_logs_from_batch(self, url_and_position: Tuple[str, int]) -> Tuple[List[Dict], str]:
+        url, line_start = url_and_position
         response = requests.request(method=GET, url=url, stream=False)
```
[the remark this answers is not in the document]
ablakley-r7: It does and we can see our content length in bytes though I'm not sure how immediately useful this is. Thinking we would have to know the compression ratio and average size of each log JSON?

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
```diff
@@ -40,21 +41,50 @@ def authenticate(self) -> None:
...
         with Pool(max_threads) as pool:
-            batch_logs = pool.imap(self.get_siem_logs_from_batch, batch_download_urls)
-            for result in batch_logs:
-                if isinstance(result, (List, Dict)):
-                    logs.extend(result)
+            result = pool.imap(self.get_siem_logs_from_batch, pool_data)
+            for batch_logs, url in result:
+                if isinstance(batch_logs, (List, Dict)):
+                    with lock:
+                        total_count.value = total_count.value + len(batch_logs)
```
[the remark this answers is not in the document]
ablakley-r7: Have added that into remove a length calculation. I think the rest is maybe fine, we only really make the necessary calculations to get our counts. The subsection of logs could be done with negative splicing but we then also have to do a min check for batches of logs that return only 1 log.

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
```diff
@@ -79,15 +109,41 @@ def get_siem_batches(
...
+            self.logger.info(f"API: Attempting to resume from last file at {saved_url}")
+            list_of_batches = list_of_batches[
+                next((index for index, url in enumerate(list_of_batches) if saved_url in url), len(list_of_batches)) :
+            ]
+
+        pool_data = [(url, starting_position if saved_url and saved_url in url else 1) for url in list_of_batches]
+        return pool_data
+
+    def get_siem_logs_from_batch(self, url_and_position: Tuple[str, int]) -> Tuple[List[Dict], str]:
+        url, line_start = url_and_position
```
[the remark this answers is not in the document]
ablakley-r7: So have refactored resume_batch and now use partial to spread the saved values into the get_siem_logs_from_batch function instead of maintaining them in a tuple that resume_batch generates.

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
```diff
@@ -79,15 +109,41 @@ def get_siem_batches(
         urls = [batch.get("url") for batch in batch_list]
         return urls, batch_response.get("@nextPage"), caught_up
 
-    def get_siem_logs_from_batch(self, url: str):
+    def resume_from_batch(
```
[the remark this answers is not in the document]
ablakley-r7: So have refactored resume_batch and now use partial to spread the saved values into the get_siem_logs_from_batch function instead of maintaining them in a tuple that resume_batch generates.

plugins/mimecast_v2/icon_mimecast_v2/tasks/monitor_siem_logs/task.py
```diff
@@ -134,7 +141,8 @@ def apply_custom_config(self, state: Dict, run_type: str, custom_config: Dict =
                 current_query_config[log_type] = {QUERY_DATE: log_query_date}
         page_size = max(1, min(custom_config.get(PAGE_SIZE, DEFAULT_PAGE_SIZE), DEFAULT_PAGE_SIZE))
         thread_count = max(1, custom_config.get(THREAD_COUNT, DEFAULT_THREAD_COUNT))
-        return page_size, thread_count
+        log_limit = custom_config.get(LOG_LIMIT)
```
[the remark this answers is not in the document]
ablakley-r7: Updated to log_limits dictionary

plugins/mimecast_v2/icon_mimecast_v2/tasks/monitor_siem_logs/task.py
```diff
@@ -134,7 +141,8 @@ def apply_custom_config(self, state: Dict, run_type: str, custom_config: Dict =
                 current_query_config[log_type] = {QUERY_DATE: log_query_date}
         page_size = max(1, min(custom_config.get(PAGE_SIZE, DEFAULT_PAGE_SIZE), DEFAULT_PAGE_SIZE))
         thread_count = max(1, custom_config.get(THREAD_COUNT, DEFAULT_THREAD_COUNT))
-        return page_size, thread_count
+        log_limit = custom_config.get(LOG_LIMITS, {})
+        return page_size, thread_count, log_limit
```
joneill-r7, line 145: minor but the typing on `apply_custom_config` needs updated to reflect returning this extra variable now

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
```diff
@@ -79,15 +115,44 @@ def get_siem_batches(
...
+        """
+        sub_list = list_of_batches[
+            next(
+                (index for index, url in enumerate(list_of_batches) if saved_url and saved_url in url),
+                len(list_of_batches),
+            ) :
+        ]
+        if sub_list:
+            return sub_list
+        return list_of_batches
```
joneill-r7, line 133: worth adding a log here to catch if we had a saved_url and it wasn't returned in the batch of URLs on the subsequent run?

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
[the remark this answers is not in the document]
joneill-r7: Sorry that didn't explain very well. I mean 'start' arg isn't jumping to that index, this should explain it a bit more from what I'm seeing:

```
>>> test = [1, 2, 4, 5]
>>> for _, line in enumerate(test, start=2):
...     print(f"index is {_}, content is {line}")
...
index is 2, content is 1
index is 3, content is 2
index is 4, content is 4
index is 5, content is 5

```

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
```diff
@@ -79,15 +115,44 @@ def get_siem_batches(
...
         with gzip.GzipFile(fileobj=BytesIO(response.content), mode="rb") as file_:
-            logs = []
             # Iterate over lines in the decompressed file, decode and load the JSON
-            for line in file_:
+            for _, line in enumerate(file_, start=line_start):
                 decoded_line = line.decode("utf-8").strip()
-                logs.append(json.loads(decoded_line))
-        return logs
+                try:
+                    logs.append(json.loads(decoded_line))
```
joneill-r7, line 146: honestly I'm a bit wary that we could opening huge gzip files here and loading the entire content into memory. do we know if there's a limit on the files from mimecast or worth looking at how we used docker stats in the past? Although it does cause issues with catching the exception, and I might be being overcautious

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
```diff
@@ -40,21 +42,55 @@ def authenticate(self) -> None:
...
+                ),
+                pool_data,
+            )
+            for batch_logs, url in result:
+                with lock:
+                    batch_logs_count = len(batch_logs)
+                    total_count.value = total_count.value + batch_logs_count
+                    if total_count.value >= log_size_limit:
+                        leftover_logs_count = total_count.value - log_size_limit
+                        subsection_of_logs = batch_logs[: (batch_logs_count - leftover_logs_count)]
```
joneill-r7, line 81: could be beneficial for memory to reassign back to batch_logs?

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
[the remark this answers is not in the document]
ablakley-r7: Ah ok, updated to just iterate through the list now we are chunking.

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
ablakley-r7: Updated to now use chunking which will keep us from storing the response.

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
```diff
@@ -79,15 +115,60 @@ def get_siem_batches(
...
+        line_start = 0
+        if saved_url and saved_url in url:
+            line_start = saved_position
+
+        response = requests.request(method="GET", url=url, stream=True)
+        logs = []
+        lines_count = 0
+        buffer = BytesIO()
+        with gzip.GzipFile(fileobj=buffer, mode="rb") as file_:
+            for chunk in response.iter_content(chunk_size=8192):
```
ekelly-r7, line 148: I'm curious, why `8192` was chosen as the `chunk_size` here?

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
```diff
@@ -79,15 +115,60 @@ def get_siem_batches(
...
+                buffer.write(chunk)
+                buffer.seek(0)
+                file_.fileobj = buffer
+                for line in file_:
+                    if lines_count < line_start:
+                        lines_count += 1
+                        continue
+                    decoded_line = line.decode("utf-8").strip()
```
ekelly-r7, line 156: Might need an outer exception here if we get some data where `.decode` or `.strip` doesn't as expected

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
```diff
@@ -79,15 +115,60 @@ def get_siem_batches(
...
+                    try:
+                        logs.append(json.loads(decoded_line))
+                    except json.JSONDecodeError:
+                        self.logger.info("API: Invalid JSON detected, skipping remainder of file")
+                        return logs, url
+
+                buffer.seek(0)
+                buffer.truncate(0)
```
ekelly-r7, line 164: Would it be possible to add a small comment as to what's happening here?

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
ablakley-r7: I've put this in one try block to catch decode errors now.

plugins/mimecast_v2/icon_mimecast_v2/util/api.py
ablakley-r7: 8192 seems reasonable and I know it is commonly used as a default buffer size.

ekelly-r7 (approved):

PR #3210, opened 2025-02-24
[SOAR-18683] Servicenow Extend OAuth with client credentials flow

lcwiklinski-r7:
https://rapid7.atlassian.net/browse/SOAR-18683

- Extend OAuth with client credentials flow
- Update the documentation
- @rmurray-r7 updated spec file for SDK automation (FedRAMP plugin)


plugins/servicenow/icon_servicenow/connection/connection.py
```diff
@@ -29,7 +29,7 @@ def connect(self, params):
         oauth_client_id = params.get(Input.CLIENT_ID)
         oauth_client_secret = params.get(Input.CLIENT_SECRET, {}).get("secretKey")
 
-        if not oauth_client_id or not oauth_client_secret:
+        if not oauth_client_id or not oauth_client_secret and username and password:
```
igorski-r7, line 32: I think we should add case where none of `username`, `password`, `client_id` and `client_secret` are provided. In that case we should raise an exception that something has to be entered.

lcwiklinski-r7: make sense, it's covered now @igorski-r7 , thanks

rmurray-r7 (approved):

rmurray-r7 (approved):

igorski-r7 (approved):

PR #3223, opened 2025-02-26
Servicenow 8.0.5 Release

lcwiklinski-r7:
https://rapid7.atlassian.net/browse/SOAR-18683

- Extend OAuth with client credentials flow
- Update the documentation
- @rmurray-r7 updated spec file for SDK automation (FedRAMP plugin)

previous PR: https://github.com/rapid7/insightconnect-plugins/pull/3210

image

### DIFF
--- a/plugins/servicenow/bin/icon_servicenow
+++ b/plugins/servicenow/bin/icon_servicenow
@@ -6,7 +6,7 @@ from sys import argv
 
 Name = "ServiceNow"
 Vendor = "rapid7"
-Version = "8.0.4"
+Version = "8.0.5"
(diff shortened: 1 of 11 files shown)


rmurray-r7 (dismissed):

plugins/servicenow/bin/icon_servicenow
```diff
@@ -6,7 +6,7 @@ from sys import argv
 
 Name = "ServiceNow"
 Vendor = "rapid7"
-Version = "8.0.4"
+Version = "8.0.5"
```
igorski-r7, line 9: Wonder if this should be major or minor version bump instead of patch as we are adding new functionality

igorski-r7 (changes requested):
LGTM, but one question 😄

plugins/servicenow/bin/icon_servicenow
lcwiklinski-r7: In my opinion, this change is considered a minor version bump because the core functionality of the authentication method remains backward-compatible. New flow introduces new capabilities without breaking or altering the existing behavior

PR #3271, opened 2025-03-05
Dig - (SDK Automation prep)

rmurray-r7:
### Description

Updating plugin so no data is lost during refresh.

A lot of the help.md was hardcoded meaning a lot of the info was lost during refresh. Unfortunately there is nowhere in the spec to ensure this data does not get lost but as it is only an input/output example it is nothing major:

image

  - SDK Bump to 6.2.5
  - spec file refactor


rmurray-r7:
VALIDATOR IS EXPECTED TO FAIL DUE TO PACKAGES. I took the packages out and refresh locally and it's fine, nothing else is breaking after the OS package check 😄 
image

plugins/dig/help.md
```diff
@@ -168,7 +141,7 @@ NXDOMAIN
 NOTIMP
 REFUSED
 YXDOMAIN
-YXRRSET
+ YXRRSET
```
joneill-r7, line 144: strange spacing in here?

rmurray-r7: Unfortunately that's done by the tooling 😞

igorski-r7 (approved):

plugins/dig/help.md
rmurray-r7: Thanks for the suggestion, I added a space then `/n` and no more rogue space before the word! 😄

PR #3272, opened 2025-03-05
Basename - (SDK Automation prep)

rmurray-r7:
### Description

Refactoring plugin for SDK automation

  - SDK Bump to 6.2.5
  - spec file refactor

### DIFF
--- a/plugins/basename/plugin.spec.yaml
+++ b/plugins/basename/plugin.spec.yaml
@@ -3,8 +3,8 @@ extension: plugin
 products: [insightconnect]
 name: basename
 title: Basename
-description: This plugin is used to get the last item of a file path or URL using
-  Python's basename utility
+description: Basename is used to get the last item of a file path or URL using Python's
+  basename utility
 version: 1.1.4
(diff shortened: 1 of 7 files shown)


rmurray-r7:
The validator is expected to fail due to a known issue within the HelpValidator. I have raised a ticket to investigate this: https://rapid7.atlassian.net/browse/SOAR-19063

plugins/basename/plugin.spec.yaml
```diff
@@ -3,42 +3,49 @@ extension: plugin
 products: [insightconnect]
 name: basename
 title: Basename
-description: This plugin is used to get the last item of a file path or URL using Python's basename utility
-version: 1.1.3
+description: This plugin is used to get the last item of a file path or URL using
```
joneill-r7, line 6: Could update the description to be "Basename is used to get the last item of a file path..." and it'll pass the validator? I think that what's the validator is saying because `Basename` isn't in this description but `basename` is (looked at a few other plugins)

PR #3281, opened 2025-03-11
[SOAR-19079] Update Linter check

rmurray-r7:
### Description

At the moment, the linter check runs different on **insightconnect-plugins** vs **komand-plugins**. This repo currently runs the check via `black . --check` whereas **komand-plugins** runs it on `CHANGED_FILES=()`. 

From testing it is clear that **insightconnect-plugins** does not detect files which require black changes due to this check whereas **komand-plugins** does. This PR will have the 2 checks in sync. I am currently making an update to the tooling which will add functionality into the refresh function to allow us to ensure we are always passing the linter without manually needing to run black and MD5.

PR for komand-plugins: https://github.com/rapid7/komand-plugins/pull/2138

  - Update Linter check

### DIFF
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -4,19 +4,19 @@ on: [push]
 
 jobs:
   lint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
 
-      - name: Set up Python 3.9
-        uses: actions/setup-python@v4
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v5
         with:
-          python-version: '3.9'
+          python-version: '3.11'
       - name: Install dependencies
         run: |
-          pip install black==22.3.0
+          pip install black==24.10.0
       - name: Lint
         run: |
           CHANGED_FILES=()
--- a/.github/workflows/black.yml
+++ b/.github/workflows/black.yml
@@ -6,17 +6,32 @@ jobs:
   lint:
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
       - name: Set up Python 3.9
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
-          python-version: 3.9
-
+          python-version: '3.9'
       - name: Install dependencies
         run: |
           pip install black==22.3.0
-
       - name: Lint
         run: |
-          black . --check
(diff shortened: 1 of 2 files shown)


.github/workflows/black.yml
```diff
@@ -6,17 +6,32 @@ jobs:
...
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
       - name: Set up Python 3.9
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
```
igorski-r7, line 14: I think we should use Python `3.11` instead of `3.9`

```suggestion
        uses: actions/setup-python@v5
```

.github/workflows/black.yml
```diff
@@ -6,17 +6,32 @@ jobs:
...
         with:
-          python-version: 3.9
-
+          python-version: '3.9'
```
igorski-r7, line 16:
```suggestion
          python-version: '3.11'
```

.github/workflows/black.yml
```diff
@@ -6,17 +6,32 @@ jobs:
   lint:
     runs-on: ubuntu-20.04
```
igorski-r7, line 7: Imo, we could bump it to `ubuntu-latest`

.github/workflows/black.yml
```diff
@@ -6,17 +6,32 @@ jobs:
...
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v3
```
igorski-r7, line 9:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/black.yml
```diff
@@ -6,17 +6,32 @@ jobs:
...
       - name: Set up Python 3.9
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
-          python-version: 3.9
-
+          python-version: '3.9'
       - name: Install dependencies
         run: |
           pip install black==22.3.0
```
igorski-r7, line 19: I don't know why we are using black version `22.3.0`. I think we could be using always the latest one 😄 

```suggestion
          pip install black
```

igorski-r7 (commented):
LGTM, I suggest updating some of the GHA workflow components to use latest versions 😄

.github/workflows/black.yml
rmurray-r7: I am going to set this to `24.10.0` as it is the version set in insight-plugin.

igorski-r7 (dismissed):

igorski-r7 (approved):

igorski-r7 (approved):

igorski-r7 (approved):

PR #3283, opened 2025-03-13
Html dev

nmccullagh-r7:
Tweak error handling


plugins/html/icon_html/util/helpers.py
```diff
@@ -17,8 +21,44 @@ def read_file_content(filename: str) -> str:
         return encode_to_base64(file_.read())
 
 
+def delete_file(file_name):
+    try:
+        os.remove(file_name)
+    except FileNotFoundError:
+        pass
+    except Exception as e:
```
dgreene-r7, line 29: Iirc best practice calls for catching the most specific exception you can rather than just `Exception` and calling it a day.

plugins/html/icon_html/util/helpers.py
```diff
@@ -17,8 +21,44 @@ def read_file_content(filename: str) -> str:
...
+    file_type: str,
+    *args,
+    **kwargs,
 ) -> str:
-    pypandoc.convert_text(input_html_string, to_format, from_format, outputfile=temporary_filename, *args, **kwargs)
-    return read_file_content(temporary_filename)
+    file_name = f"{uuid.uuid4()}_{file_type}"
+
+    try:
+        pypandoc.convert_text(
```
dgreene-r7, line 45: Maybe a stupid question but what does `convert_text` return in cases where it needs to write the temporary file?

dgreene-r7 (commented):
2 small notes. Also might be a good idea to add a unit test with an iframe that points to an external resource so we don't run into a regression somewhere down the road.

plugins/html/icon_html/util/helpers.py
nmccullagh-r7: It raises a RuntimeError exception

dgreene-r7: So returns `None`? That's a wild API choice lol.

dgreene-r7 (approved):

ekelly-r7 (approved):

PR #3301, opened 2025-03-31
Updating Values - MS Defender

rmurray-r7:
### Description

  When testing on staging, a few values were not working due to the way they were named. The docs suggested to reword them but in doing so it causes it to break 😢 Changing it works. The clean value does not work so removing as it causes plugin to break.

image


DHall-R7 (approved):

igorski-r7 (approved):

DHall-R7 (commented):
Approved

PR #3404, opened 2025-05-08
[NO-TICKET] Unit Test Coverage- Resolve deprecation of v3

rmurray-r7:
### Description

V3 of upload-artifact is being deprecated. This is preventing the unit test coverage being generated. This will update to v4

Snowballstemmer 3.0.0 is causing issues. Specifying an older version works.

https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

  - Update upload-artifact to v4 
  - Install prior version of `snowballstemmer`

### DIFF
--- a/.github/workflows/prospector.yml
+++ b/.github/workflows/prospector.yml
@@ -18,6 +18,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install prospector[with_bandit]==1.16.1
+          # Version 3.0.0 raised KeyError, this pins it to working version
           pip install snowballstemmer==2.2.0
 
       - name: Prospector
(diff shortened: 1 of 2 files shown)


.github/workflows/prospector.yml
```diff
@@ -17,7 +17,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install prospector[with_bandit]==1.9.0
+          pip install prospector[with_bandit]==1.16.1
+          pip install snowballstemmer==2.2.0
```
igorski-r7, line 21: I think we can add comment here that's the temporary solution 😃

igorski-r7 (approved):

PR #3530, opened 2025-08-12
[SOAR-19978] InsightVM - Tag Assets (Convert 'Tag Source' input to correct case)

rmurray-r7:
### Description

[Support case](https://rapid7.atlassian.net/browse/SI-30609) opened regarding the output from `Get Tag` not working for `Tag Assets`. There is an issue with the V3 API where it does not return the data correctly from UI (tag source not returned in correct case, and tag type not always correct). This is causing workflows (which use output from `Get Tag` in input for `Tag Assets`) to fail as the `Tag Assets` action requires the source to be in uppercase. I raised a [ticket](https://rapid7.atlassian.net/browse/NEX-60026) with IVM about this but we can't afford to wait til it's fixed on their end.

Realistically this could be resolved if IVM fixed their Get Tag API to return it with that the UI shows but not sure on when that can be picked up so adding this change in now.

Description also updated as I found we don't support build-in tags. I am not sure why but this API is V2 which is very old. It is not supported in V3 so we need to made do. An IDEA ticket was made to support this for V3.

  - Updated Tag Assets to convert Tag Source inputs to uppercase (as API requires it to always be uppercase)
  - SDK Bump to 6.3.9
  - Resolved Snyk bumping `aiohttp` 

### Testing

Prior to update 'custom' would not work as API requires it to be 'CUSTOM'. This is due to Get Tag not returning it in correct case. After changing it to convert source to upper it works:

Prior:
image

After:
image

### DIFF
--- a/plugins/rapid7_insightvm/.CHECKSUM
+++ b/plugins/rapid7_insightvm/.CHECKSUM
@@ -1,7 +1,7 @@
 {
-	"spec": "e69ca6b63f234dcdf3610255ef004851",
-	"manifest": "a453a4461d9bb3f033d6df20d6b51919",
-	"setup": "339a1123ce498082c330ed2d4e63bfaa",
+	"spec": "3d9931496c2cc61d03ace6f0c41bc131",
+	"manifest": "610650a75f0cbf3529548f57a5a653d6",
+	"setup": "6310a78894925f0ba3ff220d76286e2d",
 	"schemas": [
 		{
 			"identifier": "add_scan_engine_pool_engine/schema.py",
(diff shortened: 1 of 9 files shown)


simonirwin-r7:
### :tada: **Snyk checks have passed. No issues have been found so far.**

:white_check_mark: **code/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/soar-insightconnect-new/project/99b6a69d-ae8b-478a-9a48-96c3acc7f1bb/pr-check/7b5e7011-cbb6-409a-b6b0-a7dca5698e1a?refs=7fbdf82d7d2a7db39fdf46adc6003f957a09c24f&source=prChecksComment)

lcwiklinski-r7 (approved):

PR #3555, opened 2025-09-04
[SOAR-20060] Fortinet FortiGate headers authorization

lcwiklinski-r7:
https://rapid7.atlassian.net/browse/SOAR-20060

- Moved authorization from params to headers
- SDK version bump to 6.3.10

image

### DIFF
--- a/plugins/fortinet_fortigate/icon_fortinet_fortigate/util/api.py
+++ b/plugins/fortinet_fortigate/icon_fortinet_fortigate/util/api.py
@@ -1,21 +1,14 @@
-import json
 from logging import Logger
 
 import requests
-from icon_fortinet_fortigate.util.util import Helpers
-from insightconnect_plugin_runtime.exceptions import PluginException
 from requests import Session
+from insightconnect_plugin_runtime.exceptions import PluginException
+from icon_fortinet_fortigate.util.util import Helpers
+import json
 
 
 class FortigateAPI:
-    def __init__(
-        self,
-        host: str,
-        api_key: str,
-        logger: Logger,
-        authentication_type: str,
-        ssl_verify: bool = False,
-    ):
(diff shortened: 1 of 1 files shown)


simonirwin-r7:
### :tada: **Snyk checks have passed. No issues have been found so far.**

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/soar-insightconnect-new/pr-checks/8da0129f-8089-40f1-a0d9-324674e28bc9?refs=4df7f3653939f849cc3b711a5d96d634f3732b0b&source=prChecksComment)

plugins/fortinet_fortigate/icon_fortinet_fortigate/util/api.py
```diff
@@ -11,8 +11,8 @@ def __init__(self, host, api_key, logger, ssl_verify=False):
         self.logger = logger
         self.ssl_verify = ssl_verify
         self.session = Session()
-        self.session.params = {"access_token": api_key}
         self.helper = Helpers(self.logger)
+        self.headers = {"Authorization": f"Bearer {api_key}", "Content-type": "application/json"}
```
ekelly-r7, line 15: will this change not break customers who are on previous versions of FortiGate (which does support access_token in session params)?

ekelly-r7: Yeah - now I'm extra concerned about this change! 😄 

Because they enforce the Authorization header in `v7.4.5`, and we are using `v7.2.4`, means there is a risk that some customers could be also using old versions, that do not support auth header

I think we should support both (session token via params and also auth header), and bump a minor plugin version

ekelly-r7 (approved):

PR #3705, opened 2026-01-30
[DNM][SI-32667] Connectwise connection issue (#3688)

igorski-r7:
## 🎫 Ticket

Ticket: SOAR-20504

## 🧩 Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Other

## 🧠 Background & Motivation

It originally comes from the SI ticket https://rapid7.atlassian.net/browse/SI-32667, I've been able to find something in the docs here https://developer.connectwise.com/Products/ConnectWise_PSA about the different regions for the API communication (but ATM I cannot log in to precisely point out the place about it, because I cannot use the 1pass credentials.

## ✨ What Changed
- SDK bump
- plugin.spec.yaml alignment with the current approach (it was updated in 2022, last time)
- Regions support (enum) for the connection schema and handling it in the API object

## 🧪 Testing
- Adjust changes in the current unit tests
- Cannot test it in staging env, because we don't have a working sandbox at the moment
- The custom plugin was tested out on customers environment and we got a response back that everything worked fine


ablakley-r7 (dismissed):

nmccullagh-r7 (approved):

CAIT-CollinJ (commented):
This version doesn't allow for connection to an on-prem Connectwise instance since the SDP URL isn't editable in the current version. 

If we were able to change the field where region is supplied and instead of choosing a region being able to define out an SDP URL in the drop down (you can still keep the cloud region in there with their FQDN) would allow instances that are On-prem to utilize this plugin as well.

jerryac:
The documentation suggests that you can leave the Region field blank (assuming this is for on prem CW) but when you populate the settings, it auto fills the Region with NA and you cannot remove it and thus cannot proceed with it blank.  This results in the plugin attempting to establish a connection with an external Connectwise URL for routing to one of their hosted instances in AWS which fails.   The connection should stay on the local LAN and talk directly to the Connectwise PSA instance for comms.    Can you please help with understanding if this is expected behavior or a bug?  

Thanks!

PR #3814, opened 2026-03-24
SI-33125: fix token expiry, add trigger checkpointing, add filters to…

landrisek-r7:
Fixes three issues found during escalation SI-33125 where the Azure Sentinel trigger stops picking up incidents after running for a while.

What changed:

api.py — Auth token was set once in __init__ and never refreshed. After ~60 min it expires and all API calls start failing with 401. Now self.headers is a property that checks token TTL and refreshes proactively before expiry.

trigger.py — Polling window was calculated as now - interval on every tick with no state persistence. If the container restarts or lags, incidents in the gap are lost. Now uses self.state to checkpoint the last processed incident timestamp across restarts.

action.py + plugin.spec.yaml — List Incidents action had no support for status or time filters, making the Timer + List Incidents workaround useless. Added optional status and created_from params using the same generate_query_params logic the trigger uses.

Bumped version to 2.2.0.

### DIFF
--- a/plugins/azure_sentinel/icon_azure_sentinel/actions/list_incidents/action.py
+++ b/plugins/azure_sentinel/icon_azure_sentinel/actions/list_incidents/action.py
@@ -1,7 +1,9 @@
+import datetime
+
 import insightconnect_plugin_runtime
 
 from .schema import Component, Input, ListIncidentsInput, ListIncidentsOutput, Output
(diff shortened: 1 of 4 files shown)


joneill-r7:
Refresh of the plugin and this SI addressed in https://github.com/rapid7/insightconnect-plugins/pull/3829

PR #3855, opened 2026-04-23
Microsoft Log Analytics - Updated Azure API versions for Workspace Management, Shared Keys, and Saved Searches | Updated SDK to the latest version (6.5.0) | Remove shared key usage from all actions except Send Log Data

igorski-r7:
## 🎫 Ticket

Ticket: SOAR-21110

## 🧩 Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Other

## 🧠 Background & Motivation

The Microsoft Log Analytics plugin currently requires a shared key for all data access operations. This includes read-only queries, which is not necessary according to Azure Log Analytics guidelines. This requirement has caused issues for users who only need read access.

## ✨ What Changed

- Updated Azure API versions for Workspace Management, Shared Keys, and Saved Searches
- Updated SDK to the latest version (6.5.0)
- Remove shared key usage from all actions except `Send Log Data`

## 🧪 Testing

It was tested out locally.


plugins/microsoft_log_analytics/icon_microsoft_log_analytics/util/endpoints.py
```diff
@@ -25,4 +32,5 @@ class Endpoint:
         + "/subscriptions/{}/resourcegroups/{}/providers/Microsoft.OperationalInsights/workspaces/{}/savedSearches?api-version={}"
     )
 
-    API_VERSION_SAVED_SEARCH = "2020-08-01"
+    # API version constants
+    API_VERSION_SAVED_SEARCH = "2025-07-01"
```
joneill-r7, line 36: how does this work, it seems be substituted into `LIST_ALL_SEARCHES` but wondering if this is right?

plugins/microsoft_log_analytics/plugin.spec.yaml
```diff
@@ -4,10 +4,11 @@ products: [insightconnect]
 name: microsoft_log_analytics
 title: Microsoft Log Analytics
 description: Edit, run log queries with data in Azure Monitor Logs
-version: 1.2.1
+version: 1.2.2
+connection_version: 1
 vendor: rapid7
 support: rapid7
-supported_versions: ["2022-03-18"]
+supported_versions: ["2026-04-22"]
```
joneill-r7: wondering if this should match the value being passed into `API_VERSION_SAVED_SEARCH`?

plugins/microsoft_log_analytics/icon_microsoft_log_analytics/util/api.py
```diff
@@ -60,7 +66,7 @@ def _generate_signature(
             raise PluginException(preset=PluginException.Preset.INVALID_JSON)
 
     def _get_workspace_id(self, subscription_id: str, resource_group_name: str, workspace_name: str) -> None:
-        api_version = "2021-12-01-preview"
+        api_version = "2025-07-01"
```
joneill-r7: should use `API_VERSION_SAVED_SEARCH` in endpoints.py?

igorski-r7: Yup in general we could add that `API_VERSION_SAVED_SEARCH` to all of those as it looks to be shared 😃 I will update it!

plugins/microsoft_log_analytics/icon_microsoft_log_analytics/util/endpoints.py
igorski-r7: It's like previously the API version were not matching between endpoint, but for now the `2025-07-1` is shared across those endpoints.

plugins/microsoft_log_analytics/plugin.spec.yaml
igorski-r7: Good catch, I will update it! 😃

joneill-r7 (approved):

PR #3860, opened 2026-05-04
Microsoft Log Analytics - Updated Azure API versions for Workspace Ma…

igorski-r7:
## 🎫 Ticket

Ticket: SOAR-21110

## 🧩 Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Other

## 🧠 Background & Motivation

The Microsoft Log Analytics plugin currently requires a shared key for all data access operations. This includes read-only queries, which is not necessary according to Azure Log Analytics guidelines. This requirement has caused issues for users who only need read access.

## ✨ What Changed

- Updated Azure API versions for Workspace Management, Shared Keys, and Saved Searches
- Updated SDK to the latest version (6.5.0)
- Remove shared key usage from all actions except `Send Log Data`

## 🧪 Testing

It was tested out locally.


plugins/microsoft_log_analytics/icon_microsoft_log_analytics/tasks/__init__.py
lcwiklinski-r7, line 1: It doesn't seems to be used at all

plugins/microsoft_log_analytics/bin/icon_microsoft_log_analytics
```diff
@@ -23,7 +23,7 @@ def main():
             monkey.patch_all()
 
     import insightconnect_plugin_runtime
-    from icon_microsoft_log_analytics import connection, actions, triggers
+    from icon_microsoft_log_analytics import connection, actions, triggers, tasks
```
lcwiklinski-r7, line 26: from icon_microsoft_log_analytics import connection, actions, triggers

plugins/microsoft_log_analytics/icon_microsoft_log_analytics/tasks/__init__.py
igorski-r7: It's added automatically via `insight-plugin` tooling 😃

plugins/microsoft_log_analytics/bin/icon_microsoft_log_analytics
igorski-r7: Same. On refresh it added those `tasks`.

lcwiklinski-r7 (approved):

PR #3877, opened 2026-05-19
Microsoft teams app only auth

ewilson-r7:
## 🎫 Ticket

N/A

## 🧩 Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Other

## 🧠 Background & Motivation

The Microsoft Teams plugin required a dedicated user account with username/password (ROPC OAuth2 flow) to function. This created several operational problems:

- **MFA conflicts** — the service account couldn't use MFA, creating a security exception
- **Licensing costs** — a dedicated Microsoft 365 license was required for the service account
- **Security risk** — storing user credentials and bypassing MFA for automation is a compliance concern
- **Deprecated flow** — Microsoft has been discouraging ROPC and may deprecate it

This PR eliminates the user account entirely by switching to app-only authentication (client_credentials) for Graph API operations and Bot Framework for message sending.

## ✨ What Changed

**Authentication (breaking change):**
- Removed `username_password` from the connection (connection_version bumped to 8)
- Switched from ROPC (`grant_type: password`) to `client_credentials` OAuth2 flow
- Upgraded from legacy `/oauth2/token` (v1) to `/oauth2/v2.0/token` (v2) endpoint
- Uses `scope: https://graph.microsoft.com/.default` instead of `resource` parameter

**Message sending:**
- Added Bot Framework REST API integration for all send actions (`send_message`, `send_message_by_guid`, `send_html_message`)
- Messages now appear as coming from a bot identity rather than a user
- Bot authenticates via the app's own tenant (single-tenant bot support)

**API modernization:**
- Migrated all endpoints from `/beta` to `/v1.0` (GA)
- Removed spurious `{tenant_id}` from Graph API resource paths (tenant is determined by the token)
- Removed `username` input from `get_message_in_chat` action (app permissions use chat_id directly)

**Architecture refactor:**
- New `util/graph_api_client.py` — proper API client class with centralized error handling, pagination, and domain-specific helper methods
- New `util/bot_service.py` — Bot Framework service class with token management and retry logic
- New `util/constants.py` — centralized constants, timeout values, and HTTP error map
- All actions refactored to use `self.connection.client` (Graph) and `self.connection.bot` (messaging)

**Code quality:**
- All prospector issues resolved (mccabe complexity, bandit, pylint)
- SDK updated to 6.5.1
- All unit tests rewritten to mock at client level instead of requests level
- 37 tests passing

## 🧪 Testing

**Unit tests (37 passing):**
- Every action has a dedicated test class mocking the `GraphApiClient` and `BotService`
- Trigger tests verify endpoint construction, message sorting, regex compilation, and indicator extraction
- Tests validate output against generated JSON schemas

**Integration test (`insight-plugin run`):**
- Ran `send_html_message` against a live Microsoft Teams environment
- Verified full flow: Graph API auth → team lookup → channel lookup → Bot Framework auth → message delivered

**Validation:**
- `insight-plugin validate` passes all validators
- `prospector` returns 0 issues on all new util modules

### DIFF
--- a/plugins/microsoft_teams/icon_microsoft_teams/actions/add_group_owner/action.py
+++ b/plugins/microsoft_teams/icon_microsoft_teams/actions/add_group_owner/action.py
@@ -2,11 +2,6 @@ import insightconnect_plugin_runtime
 from .schema import AddGroupOwnerInput, AddGroupOwnerOutput, Input, Output, Component
 
 # Custom imports below
-from icon_microsoft_teams.util.azure_ad_utils import (
-    get_user_info,
-    get_group_id_from_name,
-    add_user_to_owners,
-)
 from insightconnect_plugin_runtime.exceptions import PluginException
 
 
@@ -20,17 +15,16 @@ class AddGroupOwner(insightconnect_plugin_runtime.Action):
         )
 
     def run(self, params={}):
-        user_id = get_user_info(self.logger, self.connection, params.get(Input.MEMBER_LOGIN))
-        if not user_id or not user_id.get("id"):
+        user = self.connection.client.get_user_info(params.get(Input.MEMBER_LOGIN))
+        user_id = user.get("id")
+
+        if not user_id:
             raise PluginException(
                 cause="The specified user does not exist.",
                 assistance="Please check that member login is correct.",
             )
-        return {
-            Output.SUCCESS: add_user_to_owners(
-                self.logger,
-                self.connection,
-                get_group_id_from_name(self.logger, self.connection, params.get(Input.GROUP_NAME)),
-                user_id.get("id"),
-            )
-        }
+
+        group_id = self.connection.client.get_group_id_from_name(params.get(Input.GROUP_NAME))
+        success = self.connection.client.add_group_owner(group_id, user_id)
+
+        return {Output.SUCCESS: success}
--- a/plugins/microsoft_teams/icon_microsoft_teams/actions/get_channels_for_team/action.py
+++ b/plugins/microsoft_teams/icon_microsoft_teams/actions/get_channels_for_team/action.py
@@ -3,6 +3,7 @@ from .schema import GetChannelsForTeamInput, GetChannelsForTeamOutput, Input, Ou
 
 # Custom imports below
 from icon_microsoft_teams.util.komand_clean_with_nulls import remove_null_and_clean
+from insightconnect_plugin_runtime.exceptions import PluginException
 
 
 class GetChannelsForTeam(insightconnect_plugin_runtime.Action):
@@ -19,6 +20,11 @@ class GetChannelsForTeam(insightconnect_plugin_runtime.Action):
         channel_name = params.get(Input.CHANNEL_NAME)
 
         teams = self.connection.client.get_teams(team_name)
+        if not teams:
+            raise PluginException(
+                cause="Team not found.",
+                assistance=f"Please verify '{team_name}' is a valid team name.",
+            )
         team_id = teams[0].get("id")
 
         channels = self.connection.client.get_channels(team_id, channel_name)
--- a/plugins/microsoft_teams/icon_microsoft_teams/actions/get_channels_for_team/action.py
+++ b/plugins/microsoft_teams/icon_microsoft_teams/actions/get_channels_for_team/action.py
@@ -2,10 +2,6 @@ import insightconnect_plugin_runtime
 from .schema import GetChannelsForTeamInput, GetChannelsForTeamOutput, Input, Output, Component
 
 # Custom imports below
-from icon_microsoft_teams.util.teams_utils import (
-    get_teams_from_microsoft,
-    get_channels_from_microsoft,
-)
 from icon_microsoft_teams.util.komand_clean_with_nulls import remove_null_and_clean
 
 
@@ -22,12 +18,11 @@ class GetChannelsForTeam(insightconnect_plugin_runtime.Action):
         team_name = params.get(Input.TEAM_NAME)
         channel_name = params.get(Input.CHANNEL_NAME)
 
-        team_array = get_teams_from_microsoft(self.logger, self.connection, team_name)
-        team = team_array[0]
+        teams = self.connection.client.get_teams(team_name)
+        team_id = teams[0].get("id")
 
-        team_id = team.get("id")
+        channels = self.connection.client.get_channels(team_id, channel_name)
 
-        channels = get_channels_from_microsoft(self.logger, self.connection, team_id, channel_name)
         clean_channels = []
         for channel in channels:
             clean_channels.append(remove_null_and_clean(channel))
--- a/plugins/microsoft_teams/icon_microsoft_teams/actions/get_reply_list/action.py
+++ b/plugins/microsoft_teams/icon_microsoft_teams/actions/get_reply_list/action.py
@@ -1,9 +1,8 @@
 import insightconnect_plugin_runtime
-from insightconnect_plugin_runtime.exceptions import PluginException
-
 from .schema import GetReplyListInput, GetReplyListOutput, Input, Output, Component
 
-from icon_microsoft_teams.util.teams_utils import get_teams_from_microsoft, get_channels_from_microsoft, get_reply_list
+# Custom imports below
+from insightconnect_plugin_runtime.exceptions import PluginException
 
 
 class GetReplyList(insightconnect_plugin_runtime.Action):
@@ -16,24 +15,26 @@ class GetReplyList(insightconnect_plugin_runtime.Action):
         )
 
     def run(self, params={}):
-        team = params.get(Input.TEAM_NAME)
-        channel = params.get(Input.CHANNEL_NAME)
+        team_name = params.get(Input.TEAM_NAME)
+        channel_name = params.get(Input.CHANNEL_NAME)
         message_id = params.get(Input.MESSAGE_ID)
 
-        team_id, channel_id = "", ""
-        teams = get_teams_from_microsoft(self.logger, self.connection, team)
-        if teams and isinstance(teams, list):
-            team_id = teams[0].get("id")
-            channels = get_channels_from_microsoft(self.logger, self.connection, team_id, channel)
-            if channels and isinstance(channels, list):
-                channel_id = channels[0].get("id")
-
-        if not team_id and not channel_id:
+        teams = self.connection.client.get_teams(team_name)
+        if not teams:
             raise PluginException(
-                cause="No team ID with channel ID was provided.",
-                assistance="Please provide the team and channel details(name or"
-                " GUID) to send the message to a specific channel.",
+                cause="Team not found.",
+                assistance="Please verify the team name is correct.",
             )
 
-        messages = get_reply_list(self.connection, team_id, channel_id, message_id)
+        team_id = teams[0].get("id")
+        channels = self.connection.client.get_channels(team_id, channel_name)
+        if not channels:
+            raise PluginException(
+                cause="Channel not found.",
+                assistance="Please verify the channel name is correct.",
+            )
+
+        channel_id = channels[0].get("id")
+        messages = self.connection.client.get_message_replies(team_id, channel_id, message_id)
+
         return {Output.MESSAGES: messages}
--- a/plugins/microsoft_teams/icon_microsoft_teams/connection/connection.py
+++ b/plugins/microsoft_teams/icon_microsoft_teams/connection/connection.py
@@ -50,20 +50,12 @@ class Connection(insightconnect_plugin_runtime.Connection):
     def test(self):
         try:
             self.client.test()
-        except PluginException as error:
-            raise ConnectionTestException(
-                cause="Graph API connection test failed.",
-                assistance="Please verify your Application ID, Directory ID, and Application Secret. "
-                "Ensure the app registration has Microsoft Graph application permissions with admin consent.",
-            ) from error
-
-        try:
             self.bot.test()
         except PluginException as error:
             raise ConnectionTestException(
-                cause="Bot Service connection test failed.",
-                assistance="Please verify your Application ID and Application Secret are correct for the Bot Service. "
-                "Ensure the bot app registration has the Bot Framework API permissions.",
+                cause=error.cause,
+                assistance=error.assistance,
+                data=error.data,
             ) from error
 
         return {"success": True}
--- a/plugins/microsoft_teams/icon_microsoft_teams/connection/connection.py
+++ b/plugins/microsoft_teams/icon_microsoft_teams/connection/connection.py
@@ -49,22 +49,21 @@ class Connection(insightconnect_plugin_runtime.Connection):
 
     def test(self):
         try:
-            self.client.authenticate()
+            self.client.test()
         except PluginException as error:
             raise ConnectionTestException(
-                cause="Unable to get authentication token.",
+                cause="Graph API connection test failed.",
                 assistance="Please verify your Application ID, Directory ID, and Application Secret. "
                 "Ensure the app registration has Microsoft Graph application permissions with admin consent.",
             ) from error
 
-        # Verify we can actually call the Graph API
         try:
-            self.client._make_request("GET", "/v1.0/organization")  # pylint: disable=protected-access
+            self.bot.test()
         except PluginException as error:
             raise ConnectionTestException(
-                cause="Authentication succeeded but API call failed.",
-                assistance="The token was obtained but could not access Microsoft Graph. "
-                "Verify the app has Organization.Read.All or similar permissions.",
+                cause="Bot Service connection test failed.",
+                assistance="Please verify your Application ID and Application Secret are correct for the Bot Service. "
+                "Ensure the bot app registration has the Bot Framework API permissions.",
             ) from error
 
         return {"success": True}
--- a/plugins/microsoft_teams/icon_microsoft_teams/connection/connection.py
+++ b/plugins/microsoft_teams/icon_microsoft_teams/connection/connection.py
@@ -3,126 +3,53 @@ from .schema import ConnectionSchema, Input
 
 # Custom imports below
 from insightconnect_plugin_runtime.exceptions import ConnectionTestException, PluginException
-import requests
-import time
 
-from icon_microsoft_teams.util.constants import TIMEOUT, RESOURCE_URL, AUTH_URL, GRAPH_SCOPE_DEFAULT
+from icon_microsoft_teams.util.constants import RESOURCE_URL
 from icon_microsoft_teams.util.graph_api_client import GraphApiClient
 from icon_microsoft_teams.util.bot_service import BotService
 
 
 class Connection(insightconnect_plugin_runtime.Connection):
     def __init__(self):
-        super(self.__class__, self).__init__(input=ConnectionSchema())
+        super().__init__(input=ConnectionSchema())
         self.client = None
         self.bot = None
-        self.app_id = None
-        self.tenant_id = None
-        self.app_secret = None
-        self.api_token = None
         self.resource_endpoint = None
 
-    def connect(self, params):
-        self.app_id = params.get(Input.APPLICATION_ID, "").strip()
-        self.tenant_id = params.get(Input.DIRECTORY_ID, "").strip()
-        self.endpoint = params.get(Input.ENDPOINT, "Normal")
-        self.app_secret = params.get(Input.APPLICATION_SECRET, {}).get("secretKey", "").strip()
+    def connect(self, params):  # pylint: disable=signature-differs
+        app_id = params.get(Input.APPLICATION_ID, "").strip()
+        tenant_id = params.get(Input.DIRECTORY_ID, "").strip()
+        endpoint = params.get(Input.ENDPOINT, "Normal")
+        app_secret = params.get(Input.APPLICATION_SECRET, {}).get("secretKey", "").strip()
 
-        self.resource_endpoint = RESOURCE_URL.get(self.endpoint)
+        self.resource_endpoint = RESOURCE_URL.get(endpoint)
 
-        # Auth tokens expire after 1 hour. Track when we last authenticated.
-        self.time_ago = 0
-        self.time_now = time.time()
-
-        # Authenticate using client_credentials (app-only)
-        self._get_app_token()
-
-        # Initialize the Graph API client
+        # Initialize the Graph API client (handles its own authentication)
         self.client = GraphApiClient(
+            app_id=app_id,
+            app_secret=app_secret,
+            tenant_id=tenant_id,
             base_url=self.resource_endpoint,
+            endpoint=endpoint,
             logger=self.logger,
-            get_headers_func=self.get_headers,
         )
 
-        # Initialize the Bot Framework service for sending messages
+        # Initialize the Bot Framework service (handles its own authentication)
         self.bot = BotService(
-            app_id=self.app_id,
-            app_secret=self.app_secret,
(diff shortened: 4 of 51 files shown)


plugins/microsoft_teams/icon_microsoft_teams/util/bot_service.py
```diff
@@ -0,0 +1,233 @@
...
+)
+
+
+class BotService:
+    """
+    Bot Framework service for sending messages to Microsoft Teams.
+
+    Uses the Bot Framework REST API to send messages as a bot identity,
+    eliminating the need for a user account. The bot must be registered
+    in Azure and installed in the target teams/channels.
```
igorski-r7, line 16: This approach of using `Azure Bot Service` the customer needs to configure their `Azure Bot` instance. We had a previously project related to the buttons and adaptive cards with usage of `Azure Bot`. It requires more work to be implemented to fully integrate `Azure Bot Instance` to properly serve events to our ChatOps approach. 

Possibly it could allow customers to get rid of the `username` and `password` credentials in OAuth `password` grant flow. They need to create their own `manifest.json` files as well as icon etc to create their own bot that matches Microsoft Bot Framework schema. 

In terms of refactoring `Send Message` actions etc to match 1:1 how its done now it looks good. The proper usage of Bot Framework endpoints has been implemented, but previously we had a conversation related to extend ChatOps functionality.

Did you test it how it works now? On our instance we do have a bot installed that I was testing previously so it can be used 😃 

From UX perspective, It would require more setup from the customers perspective on `Azure` to properly configure their Bot instance.

plugins/microsoft_teams/icon_microsoft_teams/util/bot_service.py
igorski-r7, line 1: Nice to see the separation here between Bot Service and GraphAPI client, but those two classes could inherit from some API generic class that provides methods like `_call_api()` etc.

In general we have `make_requrest` in `insightconnect_runtime.helper` methods 😃 so that it can be used as well 😃

plugins/microsoft_teams/icon_microsoft_teams/actions/add_group_owner/action.py
```diff
@@ -20,17 +15,16 @@ def __init__(self):
         )
 
     def run(self, params={}):
-        user_id = get_user_info(self.logger, self.connection, params.get(Input.MEMBER_LOGIN))
-        if not user_id or not user_id.get("id"):
+        user = self.connection.client.get_user_info(params.get(Input.MEMBER_LOGIN))
+        user_id = user.get("id")
```
igorski-r7, line 19: If we are working on that plugin update fully, would it be okay to fully refresh and refactor the plugin code to meet latest patterns? 😃 

I mean for example those plugin input bindings for variables etc.

plugins/microsoft_teams/icon_microsoft_teams/actions/get_channels_for_team/action.py
```diff
@@ -22,12 +18,11 @@ def run(self, params={}):
         team_name = params.get(Input.TEAM_NAME)
         channel_name = params.get(Input.CHANNEL_NAME)
 
-        team_array = get_teams_from_microsoft(self.logger, self.connection, team_name)
-        team = team_array[0]
+        teams = self.connection.client.get_teams(team_name)
+        team_id = teams[0].get("id")
```
igorski-r7, line 28: I think we need some guard here to check if list contain elements or not

plugins/microsoft_teams/icon_microsoft_teams/actions/get_reply_list/action.py
```diff
@@ -16,24 +15,26 @@ def __init__(self):
...
-                channel_id = channels[0].get("id")
+        teams = self.connection.client.get_teams(team_name)
+        if not teams:
+            raise PluginException(
+                cause="Team not found.",
+                assistance="Please verify the team name is correct.",
+            )
 
-        if not team_id and not channel_id:
+        team_id = teams[0].get("id")
```
igorski-r7, line 29: Yup same here, the code could be refactored in a couple of places to add more guards 😃

plugins/microsoft_teams/icon_microsoft_teams/connection/connection.py
```diff
@@ -6,121 +6,150 @@
...
+            app_id=self.app_id,
+            app_secret=self.app_secret,
+            tenant_id=self.tenant_id,
+            logger=self.logger,
+        )
+
+    def _get_app_token(self):
+        """Authenticate using OAuth2 client_credentials flow (app-only, no user account)."""
+        self.logger.info("Authenticating with client_credentials flow...")
+        token_url = f"{AUTH_URL.get(self.endpoint)}/{self.tenant_id}/oauth2/v2.0/token"
```
igorski-r7: I think in case of any refactor here not sure if we need to have authentication directly in the `connection.py`. As we have Bot Service and GraphAPI which scopes for different resources while obtaining OAuth, we probably can have authentication in that common API Client class from which GraphAPI and BotService could inherit 😃

plugins/microsoft_teams/icon_microsoft_teams/util/bot_service.py
ewilson-r7: Yes, I'm aware of the interactive button work. From what I understand this would require more than just plugin work to setup a listener to get feedback from adaptive cards. This is just a 1:1 replacement of the user actions to utilize the bot framework. I've tested in our environment with the send html message action. I've also got documentation created in order to update what we currently have since this will require additional setup on the customer side. 
Screenshot 2026-05-19 at 8 41 27 AM
[Microsoft_Teams_Plugin_Setup_Guide.md](https://github.com/user-attachments/files/28014654/Microsoft_Teams_Plugin_Setup_Guide.md)

plugins/microsoft_teams/icon_microsoft_teams/connection/connection.py
ewilson-r7: Auth logic now lives in BaseClient._authenticate(). Connection just instantiates the clients and delegates. No more token management in connection.py

plugins/microsoft_teams/icon_microsoft_teams/actions/get_channels_for_team/action.py
ewilson-r7: Added if not teams: / if not channels: checks with PluginException in get_channels_for_team, send_html_message, add_channel_to_team, remove_channel_from_team

plugins/microsoft_teams/icon_microsoft_teams/util/bot_service.py
ewilson-r7: New base_client.py with shared OAuth2 auth, _call_api(), _handle_json_response(), _raise_for_status() — both GraphApiClient and BotService inherit from it

plugins/microsoft_teams/icon_microsoft_teams/connection/connection.py
```diff
@@ -3,124 +3,68 @@
...
-                assistance="Please check your connection settings.",
-            )
-        else:
-            return {"success": True}
+                cause="Authentication succeeded but API call failed.",
+                assistance="The token was obtained but could not access Microsoft Graph. "
+                "Verify the app has Organization.Read.All or similar permissions.",
+            ) from error
+
+        return {"success": True}
```
igorski-r7, line 61: I think the connection test now should check for both GraphAPI and Bot Service if they are properly configured. Now it checks only for GraphAPI. 

I suggest creating some methods inside of the GraphAPI and Bot Service classes like `test()` so that we could have less code in here.

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
```diff
@@ -0,0 +1,185 @@
...
+        :param scope: OAuth2 scope for token request
+        :param logger: Logger instance
+        """
+        self._app_id = app_id
+        self._app_secret = app_secret
+        self._tenant_id = tenant_id
+        self._auth_url = auth_url
+        self._scope = scope
+        self._logger = logger
+        self._session = requests.Session()
```
igorski-r7: I think we had problems in the past with session objects stored inside of the class. Instead of this I suggest using pure `requests` calls with `requests.request` instead of the session 😃

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
```diff
@@ -0,0 +1,185 @@
...
+
+        try:
+            result_json = response.json()
+            self._token = result_json.get("access_token")
+        except (ValueError, KeyError) as error:
+            raise PluginException(
+                cause="Failed to parse authentication response",
+                assistance="Unexpected response from the token endpoint.",
+                data=str(error),
+            ) from error
```
igorski-r7, line 106: This all could be handled in a single `try - except`

igorski-r7:
```suggestion
                data=error,
            )
```

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
```diff
@@ -0,0 +1,185 @@
...
+                cause="Authentication request timed out",
+                assistance="Please verify network connectivity and try again.",
+                data=str(error),
+            ) from error
+        except requests.exceptions.ConnectionError as error:
+            raise PluginException(
+                cause="Unable to connect to authentication endpoint",
+                assistance=f"Could not connect to {token_url}. Please verify network connectivity.",
+                data=str(error),
+            ) from error
```
igorski-r7:
```suggestion
                data=error,
            )
```

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
```diff
@@ -0,0 +1,185 @@
...
+
+        self._token_acquired_at = time.time()
+        self._logger.info(f"Authentication successful, token: ******************{self._token[-5:]}")
+
+    def _get_token(self, force_refresh: bool = False) -> str:
+        """Get a valid access token, refreshing if expired (tokens last ~1 hour)."""
+        elapsed = time.time() - self._token_acquired_at
+        if not self._token or elapsed > 3500 or force_refresh:
```
igorski-r7: That `3500` seems to be hardcoded. In general we may have some constant value hardcoded somewhere to defaults to, but I think it could be much better to pull this timing from OAuth token response with some buffer (i.e. - 100). If not present, then defaults to some constant

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
```diff
@@ -0,0 +1,185 @@
...
+        """
+        try:
+            return self._session.request(method=method, url=url, timeout=TIMEOUT, **kwargs)
+        except requests.exceptions.Timeout as error:
+            raise PluginException(
+                cause="Request timed out",
+                assistance=f"The request to {url} timed out after {TIMEOUT} seconds. "
+                "Please verify network connectivity and try again.",
+                data=str(error),
+            ) from error
```
igorski-r7:
```suggestion
                data=error,
            )
```

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
```diff
@@ -0,0 +1,185 @@
...
+        except requests.exceptions.ConnectionError as error:
+            raise PluginException(
+                cause="Unable to connect",
+                assistance=f"Could not connect to {url}. "
+                "Please verify network connectivity and that the endpoint is correct.",
+                data=str(error),
+            ) from error
```
igorski-r7:
```suggestion
                data=error,
            )
```

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
```diff
@@ -0,0 +1,185 @@
...
+        """
+        Make an HTTP request with timeout and connection error handling.
+
+        :param method: HTTP method
+        :param url: Full URL to call
+        :param kwargs: Additional arguments passed to requests
+        :return: Response object
+        """
+        try:
+            return self._session.request(method=method, url=url, timeout=TIMEOUT, **kwargs)
```
igorski-r7: We may want to properly close the requests body here. I suggest not to use session and with pure requests it would be something like this 😃 

```python
with requests.request(...) as response
   return response
```

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
```diff
@@ -0,0 +1,185 @@
...
+            return None
+
+        try:
+            return response.json()
+        except ValueError as error:
+            raise PluginException(
+                cause="Non-JSON response received",
+                assistance="The server returned a response that could not be parsed as JSON.",
+                data=response.text[:500],
+            ) from error
```
igorski-r7, line 178: Could be integrated with `_call_api()`

plugins/microsoft_teams/icon_microsoft_teams/util/graph_api_client.py
```diff
@@ -0,0 +1,506 @@
...
+    ):
+        """
+        Initialize the Graph API client.
+
+        :param app_id: Azure App Registration client ID
+        :param app_secret: Azure App Registration client secret
+        :param tenant_id: Azure AD tenant ID
+        :param base_url: Graph API base URL (e.g., https://graph.microsoft.com)
+        :param endpoint: Endpoint type (Normal, GCC, etc.) for auth URL resolution
+        :param logger: Logger instance
```
igorski-r7, line 39: I suggest adding types to the docstrings in sphinx format and separate them by spaces for more readability. For example

```python
:param app_id: Azure App Registration client ID
:type app_id: str
...
```

plugins/microsoft_teams/icon_microsoft_teams/util/graph_api_client.py
```diff
@@ -0,0 +1,506 @@
...
+
+    def _build_teams_endpoint(self, team_name: str, explicit: bool) -> str:
+        """Build the Graph API endpoint for listing teams."""
+        if explicit and team_name:
+            parsed_team_name = urllib.parse.quote(team_name, safe="")
+            return (
+                f"/v1.0/groups?$filter=resourceProvisioningOptions/Any(x:x eq 'Team') "
+                f"and displayName eq '{parsed_team_name}'"
+            )
+        return "/v1.0/groups?$filter=resourceProvisioningOptions/Any(x:x eq 'Team')"
```
igorski-r7, line 120: I think its overengineered. It could be directly written in a simplier way and be a part of public methods 😃

plugins/microsoft_teams/icon_microsoft_teams/util/graph_api_client.py
```diff
@@ -0,0 +1,506 @@
...
+            except re.error as error:
+                raise PluginException(
+                    cause=f"Channel Name {channel_name} was an invalid regular expression.",
+                    assistance=f"Please correct {channel_name}",
+                    data=str(error),
+                ) from error
+
+        endpoint = f"/v1.0/teams/{team_id}/channels"
+        result = self._make_request("GET", endpoint)
+        channels = result.get("value", [])
```
igorski-r7, line 189: I feel like we could create some `endpoints.py` or something like that and store all necessary endpoints there.

It could be based on `str.format()` pattern or maybe a static functions that returns proper endpoints? Depends on use cases. Worth considering and talk about it 😃 

Also, suggest creating `constants.py` and move all constants there.

plugins/microsoft_teams/icon_microsoft_teams/connection/connection.py
ewilson-r7: I replaced self._session = requests.Session() and all self._session.request()/self._session.post() calls with plain requests.request()/requests.post().

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
ewilson-r7: Done. Replaced self._session with plain requests.request() and requests.post() calls throughout BaseClient. No more session state stored in the class.

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
ewilson-r7: Done. Consolidated _authenticate() into a single try-except block with specific except clauses for Timeout, ConnectionError, and ValueError/KeyError.

ewilson-r7: Keeping data=str(error) — the data parameter on PluginException expects a string for serialization in logs and API responses. Passing the raw exception object could produce inconsistent output depending on the exception type. str(error) guarantees a clean string representation. For response-based errors we already use response.text directly which is the same pattern.

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
ewilson-r7: Same as comment above

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
ewilson-r7: Addressed by the session removal. Plain requests.request() handles connection cleanup automatically — no context manager needed.

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
ewilson-r7: Keeping _authenticate() separate to avoid a circular dependency — _call_api() requires auth headers, and if auth itself called _call_api() we'd need special-case logic to skip auth on auth requests. The separate method also gives us clearer error messages specific to authentication failures vs. API call failures. Happy to discuss if you see a cleaner way to structure it.

plugins/microsoft_teams/icon_microsoft_teams/util/graph_api_client.py
ewilson-r7: Done. Updated docstrings to Sphinx format with :param:, :type:, :return:, and :rtype: annotations.

plugins/microsoft_teams/icon_microsoft_teams/util/graph_api_client.py
ewilson-r7: Simplified to a list comprehension. Kept it as a helper since it's used in two places (owners and members in create_group) — inlining would duplicate the logic.

plugins/microsoft_teams/icon_microsoft_teams/util/graph_api_client.py
ewilson-r7: We already have constants.py for static values. The API endpoints are dynamically constructed with resource IDs via f-strings (e.g. /v1.0/teams/{team_id}/channels/{channel_id}), so extracting them adds indirection without much benefit — they're most readable next to the methods that use them. If we had many static/shared endpoints I'd agree, but these are all unique per-method.

plugins/microsoft_teams/icon_microsoft_teams/util/bot_service.py
```diff
@@ -0,0 +1,158 @@
...
+            tenant_id=tenant_id,
+            auth_url=AUTH_URL.get(endpoint, "https://login.microsoftonline.com"),
+            scope=BOT_FRAMEWORK_SCOPE,
+            logger=logger,
+        )
+        self._service_url = BOT_SERVICE_URL
+
+    def test(self):
+        """Test Bot Service connectivity by authenticating."""
+        self._authenticate()
```
igorski-r7: In general you could handle `try - except` here with `PluginException` raised and passed same error data as from connection test. This could simplify and reduce number of `try - except` in connection test.

plugins/microsoft_teams/icon_microsoft_teams/util/graph_api_client.py
```diff
@@ -0,0 +1,537 @@
...
+        self._base_url = base_url
+
+    def test(self):
+        """
+        Test Graph API connectivity by authenticating and calling the organization endpoint.
+
+        :raises PluginException: If authentication or API call fails
+        """
+        self._authenticate()
+        self._make_request("GET", "/v1.0/organization")
```
igorski-r7: In general you could handle `try - except` here with `PluginException` raised and passed same error data as from connection test. This could simplify and reduce number of `try - except` in connection test.

plugins/microsoft_teams/icon_microsoft_teams/connection/connection.py
```diff
@@ -3,124 +3,67 @@
...
-                assistance="Please check your connection settings.",
-            )
-        else:
-            return {"success": True}
+                cause="Bot Service connection test failed.",
+                assistance="Please verify your Application ID and Application Secret are correct for the Bot Service. "
+                "Ensure the bot app registration has the Bot Framework API permissions.",
+            ) from error
+
+        return {"success": True}
```
igorski-r7, line 61: This could be handled in a single `try - except` for both. The error cause, assitance, data could be caught from `PluginException` and passed to `ConnectionTestException`

```python
        except PluginException as error:
            raise ConnectionTestException(cause=error.case, assistance=error.assistance, data=error.data)
```

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
igorski-r7: The error object is being handled by `str()` conversion in `insightconnect_plugin_runtime` 
(https://github.com/rapid7/komand-plugin-sdk-python/blob/master/insightconnect_plugin_runtime/exceptions.py#L187) not necessary to be same in the code 😃

ewilson-r7: Good catch! Changed all data=str(error) to data=error in exception handlers.

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
ewilson-r7: Change to get token life and fallback to 3500

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
ewilson-r7: Fixed

plugins/microsoft_teams/icon_microsoft_teams/util/base_client.py
ewilson-r7: Fixed

plugins/microsoft_teams/icon_microsoft_teams/util/bot_service.py
ewilson-r7: BotService.test() now wraps _authenticate() in a try-except and raises a PluginException with bot-specific cause/assistance if something unexpected happens.

plugins/microsoft_teams/icon_microsoft_teams/util/graph_api_client.py
ewilson-r7: Added a try-except inside GraphApiClient.test() that catches unexpected exceptions and raises a PluginException with descriptive context. PluginException itself is re-raised as-is so the connection test gets the specific error details.

plugins/microsoft_teams/icon_microsoft_teams/connection/connection.py
ewilson-r7: Collapsed the two separate try-except blocks into one.

igorski-r7 (approved):

landrisek-r7 (approved):

PR #3910, opened 2026-05-28
[ANY.RUN] Add trigger for the TI Feeds

semayellow:
## 🧩 Type of Change

- [x] Feature
- [x] Bug fix
- [ ] Other

## ✨ What Changed
* Added new trigger for the TI Feeds integration
* Fixed url submission for the Android VM
* Updated connection parameters


plugins/any_run/plugin.spec.yaml
```diff
@@ -159,6 +161,62 @@ connection:
...
     required: false
     example: WbTbwa4KFk77eQNffMJynWXm49jLwGjwPMKM9Xc4
+  ti_feeds_api_key:
+    title: API Key
+    description: API key for ANY.RUN TI Feeds
+    type: credential_secret_key
+    required: false
+    example: WbTbwa4KFk77eQNffMJynWXm49jLwGjwPMKM9Xc4
+triggers:
+  anyrun_ti_feeds_enrichment:
```
igorski-r7: I think we can be more 1:1 title matching here. 
```suggestion
  get_ti_feeds:
```

plugins/any_run/plugin.spec.yaml
```diff
@@ -705,13 +763,6 @@ actions:
           - '2 weeks'
           - 'month'
         example: 'month'
-      obj_ext_extension:
-        title: Extension
-        description: Change extension to valid
-        type: boolean
-        required: false
-        default: true
-        example: true
```
igorski-r7, line 714: Is this expected?

plugins/any_run/plugin.spec.yaml
```diff
@@ -35,13 +35,15 @@ key_features:
...
+  - Load fresh IPv4, URL, and Domain-Name indicators.
 requirements:
   - Requires an ANY.RUN Sandbox or ANY.RUN TI Lookup API key without a prefix
   - Available on ANY.RUN plans with API access, including trial
 troubleshooting:
   - Configure the connection with either an API key without a prefix.
 version_history:
-  - "3.0.0 - Migrate to the ANY.RUN SDK | Update existing actions | Add new actions | Updated SDK to the latest version (6.5.0)"
+  - "4.0.0 - Add new trigger for the TI Feeds"
+  - "3.0.0 - Migrate to the ANY.RUN SDK | Update existing actions | Add new actions"
```
igorski-r7: This needs to be reverted

```suggestion
  - "3.0.0 - Migrate to the ANY.RUN SDK | Update existing actions | Add new actions | Updated SDK to the latest version (6.5.0)"
```

plugins/any_run/plugin.spec.yaml
```diff
@@ -35,13 +35,15 @@ key_features:
...
   - Retrieve IP, Domain, URL, MD5, SHA256, SHA1 reputation.
+  - Load fresh IPv4, URL, and Domain-Name indicators.
 requirements:
   - Requires an ANY.RUN Sandbox or ANY.RUN TI Lookup API key without a prefix
   - Available on ANY.RUN plans with API access, including trial
 troubleshooting:
   - Configure the connection with either an API key without a prefix.
 version_history:
-  - "3.0.0 - Migrate to the ANY.RUN SDK | Update existing actions | Add new actions | Updated SDK to the latest version (6.5.0)"
+  - "4.0.0 - Add new trigger for the TI Feeds"
```
igorski-r7:
```suggestion
  - "4.0.0 - Added new trigger: Get TI Feeds | Updated SDK to the latest version (6.5.1)"
```

plugins/any_run/icon_any_run/triggers/anyrun_ti_feeds_enrichment/trigger.py
```diff
@@ -0,0 +1,44 @@
...
+            description=Component.DESCRIPTION,
+            input=AnyrunTiFeedsEnrichmentInput(),
+            output=AnyrunTiFeedsEnrichmentOutput(),
+        )
+
+    def run(self, params={}):
+        # START INPUT BINDING - DO NOT REMOVE - ANY INPUTS BELOW WILL UPDATE WITH YOUR PLUGIN SPEC AFTER REGENERATION
+        feed_fetch_depth = params.get(Input.FEED_FETCH_DEPTH)
+        feed_fetch_interval = params.get(Input.FEED_FETCH_INTERVAL)
+        threat_feed_access_key = params.get(Input.THREAT_FEED_ACCESS_KEY)
```
igorski-r7:
```suggestion
        feed_fetch_depth = params.get(Input.FEED_FETCH_DEPTH, 120)
        feed_fetch_interval = params.get(Input.FEED_FETCH_INTERVAL, 90)
        threat_feed_access_key = params.get(Input.THREAT_FEED_ACCESS_KEY, "")
```

plugins/any_run/icon_any_run/triggers/anyrun_ti_feeds_enrichment/trigger.py
igorski-r7, line 1: Could you please include some logging messages that could help debugging the plugin in the future? Please use `self.logger.info()` to add a log line 😃

plugins/any_run/icon_any_run/triggers/anyrun_ti_feeds_enrichment/trigger.py
```diff
@@ -0,0 +1,44 @@
...
+                urls = get_indicators(connector, "url", feed_fetch_depth)
+
+            self.send(
+                {
+                    Output.ANYRUN_FEED_DOMAINS: domains,
+                    Output.ANYRUN_FEED_IPS: ips,
+                    Output.ANYRUN_FEED_URLS: urls,
+                    Output.THREAT_FEED_ACCESS_KEY: threat_feed_access_key,
+                }
+            )
```
igorski-r7: The `self.send()` method should be executed only in case we would like to trigger the trigger 😃 Currently in your code, it would send the data back from trigger on each iteration.

Please double check that behavior and add proper conditions for `domains`, `ips`, `urls` so that when no data is available it won't trigger the workflow.

semayellow: > The `self.send()` method should be executed only in case we would like to trigger the trigger 😃 Currently in your code, it would send the data back from trigger on each iteration.
> 
> Please double check that behavior and add proper conditions for `domains`, `ips`, `urls` so that when no data is available it won't trigger the workflow.

Hi @igorski-r7 
The trigger is designed to activate when new indicators are received and send them to the workflow to update the threatfeed in the SIEM. I implemented similar logic to [the existing trigger](https://github.com/rapid7/insightconnect-plugins/blob/9e99f01abe0db5bbe457219c63b9133017309d08/plugins/aws_securityhub/icon_aws_securityhub/triggers/sqs_feed/trigger.py#L59-L94). So, please let me know if I did something wrong.

plugins/any_run/plugin.spec.yaml
```diff
@@ -24,7 +24,7 @@ hub_tags:
   features: []
 sdk:
   type: slim
-  version: 6.5.0
+  version: 6.5.1
```
igorski-r7: We have deployed new version of the SDK. Would be great if you could set it to `6.6.0` 😃 Also, please refresh the plugin when it gets updated. Thanks!

```suggestion
  version: 6.6.0
```

plugins/any_run/icon_any_run/triggers/get_ti_feeds/trigger.py
```diff
@@ -0,0 +1,58 @@
...
+                self.logger.info(f"[ANY.RUN] Authentication is passed.")
+
+                domains = get_indicators(connector, "domain", feed_fetch_depth)
+                ips = get_indicators(connector, "ip", feed_fetch_depth)
+                urls = get_indicators(connector, "url", feed_fetch_depth)
+                self.logger.info(
+                    f"[ANY.RUN] TI Feeds are fetched.\nReceived:"
+                    f" {len(domains) if domains else 0} domains,"
+                    f" {len(ips) if ips else 0} ips,"
+                    f" {len(urls) if urls else 0} urls."
```
igorski-r7: As `get_indicators()` function returns `List` always it's not necessary to have an inline if-else statement here 😃 


```suggestion
                    f" {len(domains)} domains,"
                    f" {len(ips)} ips,"
                    f" {len(urls)} urls."
```

plugins/any_run/icon_any_run/triggers/get_ti_feeds/trigger.py
```diff
@@ -0,0 +1,58 @@
...
+        # START INPUT BINDING - DO NOT REMOVE - ANY INPUTS BELOW WILL UPDATE WITH YOUR PLUGIN SPEC AFTER REGENERATION
+        feed_fetch_depth = params.get(Input.FEED_FETCH_DEPTH, 120)
+        feed_fetch_interval = params.get(Input.FEED_FETCH_INTERVAL, 90)
+        threat_feed_access_key = params.get(Input.THREAT_FEED_ACCESS_KEY, "")
+        # END INPUT BINDING - DO NOT REMOVE
+
+        while True:
+            self.logger.info(f"[ANY.RUN] Initialized TI Feeds enrichment.")
+            with FeedsConnector(self.connection.feeds_api_key, integration=Config.VERSION) as connector:
+                connector.check_authorization()
```
igorski-r7: Is it required? I mean, it just checks if credentials are fine. However, not sure if we need that here, as `get_indicators()` if not properly authorized should raise an exception as well 😃

plugins/any_run/icon_any_run/triggers/get_ti_feeds/trigger.py
```diff
@@ -0,0 +1,58 @@
...
+                self.logger.info(f"[ANY.RUN] Authentication is passed.")
+
+                domains = get_indicators(connector, "domain", feed_fetch_depth)
+                ips = get_indicators(connector, "ip", feed_fetch_depth)
+                urls = get_indicators(connector, "url", feed_fetch_depth)
```
igorski-r7: I suggest wrapping it with `try/except` and raising `PluginException` from `insightconnect-plugin-runtime` to properly handle error messaging.

plugins/any_run/Dockerfile
```diff
@@ -1,34 +1,34 @@
-FROM --platform=linux/amd64 rapid7/insightconnect-python-3-slim-plugin:6.5.0 AS builder
+FROM --platform=linux/amd64 rapid7/insightconnect-python-3-slim-plugin:6.6.0 AS builder
 
-WORKDIR /workspace
+WORKDIR /python/src
```
igorski-r7: As I can see it's reverting back the workdir to `/python/src`. It's due to you are not probably using the latest version of `insight-plugin`. I suggest updating `insight-plugin` to latest version and refresh it 😃

igorski-r7 (approved):

PR #3931, opened 2026-06-03
Add installed_apps support to create_teams_chat action

ewilson-r7:
## 🎫 Ticket

N/A

## 🧩 Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Other

## 🧠 Background & Motivation

The Microsoft Teams plugin's bot uses the Bot Framework REST API to send messages to chats and channels. However, bots cannot be added as regular `aadUserConversationMember` entries — Microsoft Graph requires them to be installed as apps via the `installedApps` endpoint. Without this, any attempt to send a bot message to a chat it hasn't been added to returns a 403 Forbidden.

Previously there was no way for the plugin to automatically handle this — users had to manually add the bot to each chat before sending messages.

## ✨ What Changed

**New connection field — `app_catalog_id` (optional):**
- When configured, the plugin automatically installs the bot into chats on first message send
- Eliminates the need for manual bot installation per chat or additional workflow steps
- Found in Teams Admin Center → Manage Apps → Your bot app → App ID

**Auto-install behavior in `send_chat_message`:**
- If the bot gets a 403 (not authorized) and `app_catalog_id` is configured, it automatically installs itself into the chat via `POST /chats/{id}/installedApps` and retries the send
- If `app_catalog_id` is not configured, raises a clear error message guiding the user
- No behavior change for channel messages (admin installs bot in teams manually via manifest)

Screenshot 2026-06-08 at 8 44 58 AM


plugins/microsoft_teams/plugin.spec.yaml
```diff
@@ -1244,3 +1250,35 @@ actions:
...
+        required: true
+        order: 3
+        example: 05F59CEC-A742-4A50-A62E-202A57E478A4
+    output:
+      success:
+        title: Success
+        description: Boolean indicating if the app installation was successful
+        type: boolean
+        required: true
+        example: true
```
igorski-r7: I am unsure about having it as a separate action as it could potentially increase complexity for users?

For me, if bot is not installed within the selected Chat or Team, it should automatically add itself or create chat on `Send Message`?

For example the flow could be to first check if bot is added in Chat or Team, and based on that it would determine whether it needs to be added or not? 😃

ewilson-r7: Good point about user experience. I considered the auto-install approach but opted against it for a few reasons:

Permission scope creep — auto-installing would require TeamsAppInstallation.ReadWrite* and AppCatalog.Read.All permissions on the Send Message action, which currently needs zero Graph permissions (it uses Bot Framework only). Users who already have their bot installed would be forced to grant broader permissions for no benefit.

App Catalog ID discovery — the plugin knows its App Registration client ID but not its Teams App Catalog ID. Auto-install would need an extra lookup call on every send (or at least on failure), adding latency and another permission or we'd have to add the ID as input to the connection or message actions.

Enterprise governance — in many orgs, app installation is a controlled administrative action. Silently installing the bot could conflict with compliance policies.

The install_app action is intended as a one-time setup step (or part of a provisioning workflow), not something users call repeatedly. The create_teams_chat action already supports installed_apps for the "create + install in one shot" use case.

That said, if we want to reduce friction, we could add better error messaging on the 403 — something like "Bot is not installed in this team/chat. Use the Install App action to install it." That way users get a clear next step without implicit side effects.

igorski-r7: I am still up to improve the users experience and this could be automated 😃 I mean, we don't need action to install application on teams, as customers manually create their bot `manifest.json` file. Saying that, they can just upload and installed Bot manually (admins) on specific Teams instance.

Based on that, we won't require any additional permissions like `TeamsAppInstallation.ReadWrite*`, as we won't handle it. It will be fully done as a part of plugin connection, so the customer needs to manually install the Bot while creating its manifest etc. 😃 

I am still like we don't need that action. We can manually handle adding that bot to the Chat, or to manually create a chat with user depending on `Send Message` action input fields. The flow still would be to check if bot is in channel, if not, let's get added to that channel and send message. Similar for `Send Message` to the user. Bot needs to create chat with specific user at first (if there was no chat perviously 😃 ) and then send message.

I know it requires some permissions, but for proper UX we need to have it. In that way, we would maintain current plugin behaviour with no need to add additional steps per installing the bot app.

ewilson-r7: Agreed. I've reworked this based on your feedback. I removed the install_app action entirely and also dropped the installed_apps input from create_teams_chat since it's no longer needed.

Instead, I added an optional app_catalog_id field to the connection that users configure once during setup. When set, the bot service will automatically install itself into a chat on the first message send if it gets a 403. It installs the app, retries the send, and the message goes through — no extra actions or workflow steps required.

The flow for users is now: admin uploads the bot manifest and installs it in teams for channel messaging (same as before), then sets the App Catalog ID in the plugin connection. After that, sending messages to chats just works transparently. The only additional permission needed is TeamsAppInstallation.ReadWriteForChat.All, and only if they configure the App Catalog ID.

plugins/microsoft_teams/plugin.spec.yaml
```diff
@@ -584,6 +584,12 @@ connection:
     - GCC
     - GCC High
     - DoD
+  app_catalog_id:
```
ekelly-r7, line 587: this file needs to be updated to reflect the changes and run the refresh

ekelly-r7 (changes requested):

plugins/microsoft_teams/plugin.spec.yaml
joneill-r7: agree with Emmet's comment here. from a quick glance at the changes as well this shouldn't be a major bump (from the release branch name)

ewilson-r7: Updated change log to reflect this addition. Are you saying this shouldn't be a major release change? I did remove username/password from the connection in the release branch.

joneill-r7: ok, I didn't know this was building on top of a previous PR. In that case the major does sense if the connection has now removed fields

plugins/microsoft_teams/icon_microsoft_teams/util/bot_service.py
```diff
@@ -116,14 +128,30 @@ def send_chat_message(self, chat_id: str, message: str, content_type: str = "tex
             "textFormat": "plain" if content_type == "text" else "xml",
         }
 
-        return self._send_activity(endpoint, activity)
+        response = self._send_activity_raw(endpoint, activity)
+
+        if response.status_code == 403 and self._app_catalog_id and self._graph_client:
+            self._logger.info("Bot not in chat — auto-installing via App Catalog ID and retrying...")
+            self._graph_client.install_app_in_chat(chat_id, self._app_catalog_id)
```
joneill-r7: it seems a bit clunky, that we have these two clients for the bot and graph service, but then assign this graph client into the bot. And only for a single use case as well? Is this better handled from the action level, if we get a 403, to then install the bot?

ewilson-r7: Fair point. Passing the graph client into the bot service is a bit of a coupling smell, and you're right that it's only used for this one case. I think moving the retry logic to the action level is cleaner architecturally.

plugins/microsoft_teams/icon_microsoft_teams/actions/send_message/action.py
```diff
@@ -61,3 +61,22 @@ def run(self, params={}):
...
+        Send a chat message, auto-installing the bot if it gets a 403 and
+        app_catalog_id is configured on the connection.
+        """
+        try:
+            return self.connection.bot.send_chat_message(chat_id, message)
+        except PluginException as error:
+            if "not authorized" not in error.cause:
+                raise
+
+            app_catalog_id = self.connection.app_catalog_id
```
joneill-r7: if we have app_catalog_id as part of `GraphApiClient` these could be simplified? 

```python
if not self.connection.client.app_catalog_id:
    raise

self.logger.info(....)
self.connection.client.install_app_in_chat(chat_id)
```
That way we can just access directly the catalog ID when installing in the payload? Before this commit i think you had it as part of the bot client?

ewilson-r7: Good call. Moved app_catalog_id into GraphApiClient so install_app_in_chat uses it internally and no longer needs it passed as a parameter. The action code now just checks self.connection.client.app_catalog_id and calls self.connection.client.install_app_in_chat(chat_id) directly, keeping the bot service completely decoupled.

plugins/microsoft_teams/unit_test/test_bot_service.py
```diff
@@ -0,0 +1,85 @@
...
+        mock_response.status_code = 200
+        mock_response.json.return_value = {"id": "msg-003"}
+        self.bot._call_api.return_value = mock_response
+
+        result = self.bot.send_chat_message("19:chat-id@thread.v2", "<b>Hello</b>", content_type="html")
+        self.assertEqual(result["id"], "msg-003")
+
+        call_kwargs = self.bot._call_api.call_args
+        activity = call_kwargs.kwargs.get("json") or call_kwargs[1].get("json")
+        self.assertEqual(activity["textFormat"], "xml")
```
joneill-r7, line 70: is this right that we send html and it sets the textFormat to xml? I'm maybe just missing some context around this

joneill-r7 (commented):
one question around a unit test assertion

plugins/microsoft_teams/unit_test/test_bot_service.py
ewilson-r7: Yes, that's correct. The Bot Framework REST API uses textFormat: "xml" to indicate rich/HTML content, not "html". It's a Bot Framework convention where "xml" means "parse this as markup" (which includes HTML tags), and "plain" means plain text. A bit confusing but that's how their schema works.

plugins/microsoft_teams/plugin.spec.yaml
```diff
@@ -584,6 +584,12 @@ connection:
...
+    title: App Catalog ID
+    description: The Teams App Catalog ID of the bot. When set, the plugin will automatically install the bot into chats before sending messages. Find this in the Teams Admin Center under Manage Apps.
+    type: string
+    required: false
+    example: 05F59CEC-A742-4A50-A62E-202A57E478A4
```
igorski-r7, line 592: I wonder if we need that `app_catalog_id`? I mean, not 100% sure, but I think `conversations` endpoint from Bot API allows to create conversation with user/channel, so that if conversation is created we can simply send message based on conversation id? 😃 

In that case, we wouldn't need any additional connection parameter or something. Those conversations can also be listed via `GET` endpoint and it could be automatically done on the action's code side.

ewilson-r7: I looked into this. The Bot Framework `POST /v3/conversations` endpoint can create a 1:1 conversation between the bot and a user, but it still requires the bot app to be installed in the user's personal scope or in a team they belong to. The Microsoft docs explicitly state that if the bot isn't installed, you get a 403 `ForbiddenOperationException`. So the Create Conversation API handles "start a new conversation with a known user" but doesn't solve the "bot isn't installed yet" problem. The `app_catalog_id` handles that prerequisite step. That said, if we wanted to support proactive messaging to individual users (not existing group chats), we could combine both: use `app_catalog_id` to install the bot for the user first, then use Create Conversation to initiate the chat. But for our current use case of sending to existing chats, the install-on-403 approach covers it.

igorski-r7 (approved):

PR #3946, opened 2026-06-09
Rename to Microsoft Entra ID Admin and add Get User Memberships action

ewilson-r7:
## 🎫 Ticket

N/A

## 🧩 Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Other

## 🧠 Background & Motivation

Microsoft renamed Azure Active Directory to Microsoft Entra ID in July 2023. The plugin title needed to be updated to reflect the current product name. Additionally, there was no action to retrieve the groups and directory roles a specific user belongs to, which is a common lookup in identity management workflows.

## ✨ What Changed

- Renamed plugin title from "Azure AD Admin" to "Microsoft Entra ID Admin" (folder and package names unchanged)
- Updated description, links, tags, and keywords to reference Entra ID
- Added new `get_user_memberships` action that calls the Microsoft Graph `/users/{user_id}/memberOf` endpoint
  - Returns all groups and directory roles a user is a direct member of
  - Supports pagination for users with many memberships
  - Output typed as `[]object` to accommodate mixed types (groups, directory roles, admin units)
- Bumped version to 5.1.0

## 🧪 Testing

- Unit tests added for the new action covering:
  - Happy path: valid user returns memberships and count
  - Error case: invalid user returns 404 with appropriate PluginException
- Full existing test suite passes (30 tests, no regressions)
- Ran locally with success(see screenshot below)
Screenshot 2026-06-09 at 10 35 56 AM


plugins/azure_ad_admin/icon_azure_ad_admin/actions/get_user_memberships/action.py
```diff
@@ -0,0 +1,56 @@
...
+        user_id = params.get(Input.USER_ID)
+        # END INPUT BINDING - DO NOT REMOVE
+
+        headers = self.connection.get_headers(self.connection.get_auth_token())
+        headers["ConsistencyLevel"] = "eventual"
+
+        url = Endpoint.USER_MEMBER_OF.format(self.connection.tenant, user_id=user_id)
+
+        memberships = []
+        for _ in range(1_000):
```
joneill-r7: can we get this loop iteration done within a loop step type in ICON? this plugin can also run in cloud and those are capped at a much shorter run time than those on an orch. I'd suggest taking the `@odata.nextLink` as an optional input parameter?

plugins/azure_ad_admin/unit_test/test_get_user_memberships.py
```diff
@@ -0,0 +1,55 @@
...
+    @classmethod
+    @patch("requests.post", side_effect=Util.mocked_requests)
+    def setUpClass(cls, mock_request: MagicMock) -> None:
+        cls.action = Util.default_connector(GetUserMemberships())
+
+    @parameterized.expand(
+        [
+            [
+                "valid_user",
+                Util.read_file_to_dict("inputs/get_user_memberships.json.inp"),
```
joneill-r7: nitpit: I personally dislike using these `read_file_to_dict` for such small inputs like this, I think it's easier to just include it inline with a `json.dumps({"user_id": "user@example.com"})`. It just saves an extra click to see what it being passed here. but not a blocking comment

plugins/azure_ad_admin/unit_test/test_get_user_memberships.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@patch("requests.request", side_effect=Util.mocked_requests)
+class TestGetUserMemberships(TestCase):
+    @classmethod
+    @patch("requests.post", side_effect=Util.mocked_requests)
+    def setUpClass(cls, mock_request: MagicMock) -> None:
+        cls.action = Util.default_connector(GetUserMemberships())
+
+    @parameterized.expand(
```
joneill-r7: we don't really need these parametrised expands as we're only passing one set of parameters into the test. But it might be beneficial after implementing the `@odata.nextLink` in the response for these happy path test block.  (comment around how we can't set a loop to 1000 iterations on cloud)

plugins/azure_ad_admin/icon_azure_ad_admin/actions/get_user_memberships/action.py
ewilson-r7: Good call. I've reworked this to remove the loop entirely. The action now fetches a single page and returns a next_link output when more results are available. Users can drive pagination through a loop step in the workflow builder. Also updated the action description to note the 100-per-page default from the Graph API.

plugins/azure_ad_admin/unit_test/test_get_user_memberships.py
ewilson-r7: Agreed, switched to inline inputs for these simple single-field cases. Much easier to read at a glance.

plugins/azure_ad_admin/unit_test/test_get_user_memberships.py
ewilson-r7: Removed the parameterized expand since we're only passing one set of params per test. Added a separate test method for the next_link pagination path instead.

plugins/azure_ad_admin/icon_azure_ad_admin/actions/get_user_memberships/action.py
```diff
@@ -0,0 +1,62 @@
...
+        self.logger.info(f"Found {member_count} memberships for user {user_id}.")
+
+        if output_next_link:
+            self.logger.info("Additional pages available via next_link.")
+
+        return {
+            Output.MEMBERSHIPS: clean(memberships),
+            Output.COUNT: member_count,
+            Output.NEXT_LINK: output_next_link,
+        }
```
joneill-r7: are we better running `clean` on this entire return so that the `next_link` is removed entirely when no extra pages are available?

plugins/azure_ad_admin/icon_azure_ad_admin/actions/get_user_memberships/action.py
```diff
@@ -0,0 +1,62 @@
...
+        # END INPUT BINDING - DO NOT REMOVE
+
+        headers = self.connection.get_headers(self.connection.get_auth_token())
+        headers["ConsistencyLevel"] = "eventual"
+
+        # Use next_link if provided, otherwise start from the beginning
+        if next_link:
+            url = next_link
+        else:
+            url = Endpoint.USER_MEMBER_OF.format(self.connection.tenant, user_id=user_id)
```
joneill-r7, line 35: minor but should we log here just for traceability that we're using next link?

perhaps a logger as well of grabbing `result.get("count")`, to show how many iterations are expected?

plugins/azure_ad_admin/icon_azure_ad_admin/actions/get_user_memberships/action.py
```diff
@@ -0,0 +1,62 @@
...
+        except ValueError:
+            raise PluginException(preset=PluginException.Preset.INVALID_JSON)
+
+        memberships = result.get("value", [])
+        output_next_link = result.get("@odata.nextLink", "")
+
+        member_count = len(memberships)
+        self.logger.info(f"Found {member_count} memberships for user {user_id}.")
+
+        if output_next_link:
```
joneill-r7: minor: this could be tidied to as we don't need the defaults `[]` / `""` and then slightly less lines for the if condition as:

```python
memberships = result.get("value")
if output_next_link := result.get("@odata.nextLink"):
    self.logger.info("Additional pages...")
```

joneill-r7 (commented):
couple of extra minor comments just to help any debugging once this goes live / code tidy ups

plugins/azure_ad_admin/icon_azure_ad_admin/actions/get_user_memberships/action.py
ewilson-r7: Good idea — wrapped clean() around the entire return dict now so next_link gets stripped from the output entirely when there's no next page. Cleaner for anyone consuming the output downstream.

plugins/azure_ad_admin/icon_azure_ad_admin/actions/get_user_memberships/action.py
ewilson-r7: Added a log when using a provided next_link so it's clear in the logs we're paginating rather than starting fresh. Also logging @odata.count when the API returns it so you can see how many total memberships are expected across all pages.

plugins/azure_ad_admin/icon_azure_ad_admin/actions/get_user_memberships/action.py
ewilson-r7: Cleaned up — using the walrus operator for the next_link assignment and dropped the unnecessary defaults. Reads nicer now, thanks.

ekelly-r7 (approved):

joneill-r7 (approved):

PR #3949, opened 2026-06-11
Automox chunk CVEs and Update Example

ewilson-r7:
## 🎫 Ticket

N/A

## 🧩 Type of Change

- [x] Bug fix

## 🧠 Background & Motivation

The Automox `submit_remediation` action was failing with a 400 error from the API when devices contained more than 500 CVEs. The API enforces a per-device limit of <=500 CVEs per request, but the plugin only chunked by device count (100 devices per request) without accounting for CVE list size. This caused the entire request to fail for organizations with heavily vulnerable devices.

## ✨ What Changed

- Added CVE sub-chunking logic to `submit_remediation` in `api_client.py`. Devices with >500 CVEs are now split into multiple entries (same device ID, CVE lists capped at 500) before the device-count chunking occurs.
- Extracted `_parse_and_validate_devices()` and `_expand_devices_by_cve_limit()` helper methods to reduce cyclomatic complexity of `submit_remediation` (was 11, now under 10).
- Added `REMEDIATION_MAX_CVES_PER_DEVICE = 500` class constant.

## 🧪 Testing

- All 9 existing unit tests pass (single chunk, multi-chunk 250 devices, exactly 100 devices, invalid inputs, API error propagation).
- Ran `insight-plugin run tests/submit_remediation.json` against live API with 1044 real devices (220 exceeding 500 CVEs, max 1332). After the fix, devices expand to 1348 entries across 14 chunked requests, all returning `{"status": "accepted"}`.
- Verified mccabe complexity is now under the threshold (`submit_remediation` no longer flagged by MC0001).
- `insight-plugin validate` passes cleanly.

### DIFF
--- a/plugins/automox/icon_automox/util/api_client.py
+++ b/plugins/automox/icon_automox/util/api_client.py
@@ -454,8 +454,8 @@ class ApiClient:
         return self.remove_null_values(self._call_api("GET", f"{self.endpoint}/events", params))
 
     # Remediations
-    @staticmethod
-    def _validate_remediation_device(index: int, device: Dict) -> None:
+    def _validate_and_expand_device(self, index: int, device: Dict) -> List[Dict]:
+        """Validate a device entry and split into multiple entries if CVEs exceed the API limit."""
         if not isinstance(device.get("id"), str):
             raise PluginException(
                 cause="Invalid input",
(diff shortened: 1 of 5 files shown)


plugins/automox/icon_automox/util/api_client.py
```diff
@@ -466,15 +467,8 @@ def _validate_remediation_device(index: int, device: Dict) -> None:
...
-    def submit_remediation(self, org_id: int, action_type: str, devices_input: str) -> Dict:
-        """
-        Submit device and CVE data for remediation or matching.
-        Automatically chunks large payloads into batches of 100 devices.
-        :param org_id: Organization ID
-        :param action_type: 'remediate' or 'match'
-        :param devices_input: JSON string containing an array of device objects
-        :return: Dict with batch_uuid, total_devices, chunks_sent, and collected responses
-        """
+    def _parse_and_validate_devices(self, action_type: str, devices_input: str) -> List[Dict]:
```
joneill-r7: I know it's not part of this PR, but I think our SDK schema validation already checks this action_type matches the enum values and would never get this far to get validated so it's dead code.

If you build the plugin image and try running this action with an invalid 'action_type' the SDK should raise this error

plugins/automox/icon_automox/util/api_client.py
```diff
@@ -498,6 +492,38 @@ def submit_remediation(self, org_id: int, action_type: str, devices_input: str)
         for i, device in enumerate(devices):
             self._validate_remediation_device(i, device)
 
+        return devices
+
+    def _expand_devices_by_cve_limit(self, devices: List[Dict]) -> List[Dict]:
+        """Split devices with more than REMEDIATION_MAX_CVES_PER_DEVICE CVEs into multiple entries."""
+        expanded_devices = []
+        max_cves = self.REMEDIATION_MAX_CVES_PER_DEVICE
+        for device in devices:
```
joneill-r7: we're already iterating devices and checking that CVEs are provided. I think this validation on the max number of CVEs could be implemented at that point? 
https://github.com/rapid7/insightconnect-plugins/blob/cffd9df92631bf4d7a46dc12e770bbe117b22c96/plugins/automox/icon_automox/util/api_client.py#L464

plugins/automox/icon_automox/util/api_client.py
ewilson-r7: Good catch, thanks for the review. I removed the dead action_type validation since the SDK schema handles that upstream.

plugins/automox/icon_automox/util/api_client.py
ewilson-r7: I merged the CVE expansion logic into the validation loop so we only iterate the device list once.

igorski-r7 (approved):

PR #3951, opened 2026-06-16
[SOAR-21458] Fix Get User Info validation when manager has employeeOrgData set

lcwiklinski-r7:
## 🎫 Ticket
https://rapid7.atlassian.net/browse/SOAR-21458

## 🧩 Type of Change

- [ ] Feature
- [ ✅ ] Bug fix
- [ ] Other

## 🧠 Background & Motivation
Customer reported that `get_user_info` failed with `jsonschema.ValidationError` when the target user's manager had employeeOrgData populated:

`{'costCenter': '166561506'} is not of type 'string'`

The output schema typed `manager.employeeOrgData` as a string, but per Microsoft Graph it is a complex type with `costCenter` and `division`.

## ✨ What Changed
- plugin.spec.yaml: defined new `employeeOrgData` type (costCenter, division) and re-typed manager.employeeOrgData to use it
- Bumped version to `6.0.0`
- Unit tests: added populated employeeOrgData to the manager mock and updated expected_response so the existing test guards against regression.

## 🧪 Testing
- Unit tests
- Manual testing steps

### DIFF
--- a/plugins/azure_ad_admin/bin/icon_azure_ad_admin
+++ b/plugins/azure_ad_admin/bin/icon_azure_ad_admin
@@ -6,7 +6,7 @@ from sys import argv
 
 Name = "Microsoft Entra ID Admin"
 Vendor = "rapid7"
-Version = "5.1.1"
+Version = "6.0.0"
 Description = "Microsoft Entra ID Admin performs administrative tasks in Microsoft Entra ID (formerly Azure AD).It uses the [User](https://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0) endpoint in the [Microsoft Graph API](https://docs.microsoft.com/en-us/graph/overview?view=graph-rest-1.0)"
 
 
--- a/plugins/azure_ad_admin/bin/icon_azure_ad_admin
+++ b/plugins/azure_ad_admin/bin/icon_azure_ad_admin
@@ -6,7 +6,7 @@ from sys import argv
 
 Name = "Microsoft Entra ID Admin"
 Vendor = "rapid7"
-Version = "5.1.0"
+Version = "5.1.1"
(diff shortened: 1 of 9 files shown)


plugins/azure_ad_admin/bin/icon_azure_ad_admin
```diff
@@ -6,7 +6,7 @@ from sys import argv
 
 Name = "Microsoft Entra ID Admin"
 Vendor = "rapid7"
-Version = "5.1.0"
+Version = "6.0.0"
```
ekelly-r7, line 9: seems a bit excessive to do a major version bump for an issue that only a single customer is hitting (sounds like a quirk with their environment) 

is it possible to handle the edge case in the code, instead of doing a major version breaking change?

ekelly-r7 (changes requested):

plugins/azure_ad_admin/bin/icon_azure_ad_admin
lcwiklinski-r7: yeah, but the output schema changes, and the validator is forcing me to bump the major version.

lcwiklinski-r7:
@joneill-r7 I've been able to reproduce it 
image

igorski-r7 (approved):

igorski-r7 (changes requested):
Is this issue only for a single customer? Do we need to update the response structure that only one customer is hitting? Would it affect others in that case?

lcwiklinski-r7:
> Is this issue only for a single customer? Do we need to update the response structure that only one customer is hitting? Would it affect others in that case?

It won't affect the others - I've been checking both scenarios with the new ouput - it's optional @igorski-r7

joneill-r7 (approved):

ekelly-r7 (approved):

igorski-r7 (approved):

PR #3952, opened 2026-06-17
[SOAR-21326] insight vm scan completion fix

lcwiklinski-r7:
## 🎫 Ticket
https://rapid7.atlassian.net/browse/SOAR-21326

## 🧩 Type of Change

- [ ] Feature
- [✅ ] Bug fix
- [ ] Other

## 🧠 Background & Motivation
The `scan_completion` trigger polls IVM for finished scans and emits an event with an SQL report for each one. A few production issues on IVM 8.x consoles broke this:

- When several scans finished between two polls, only the most recent one got an event -- the rest were silently dropped.
- Report download failed with HTTP 406. IVM 8.x rejects Accept: application/json on the SQL report endpoint because the response body is CSV.
- Long reports (tens of minutes) sometimes lost their keep-alive connection during status polling. A single RemoteDisconnected killed the whole run.
- Agent scans got picked up by the trigger, but IVM refuses to scope SQL reports to them, so the run crashed.

## ✨ What Changed
Split into 5 small commits to make review easier:

- `Scan Completion: don't drop scans completed between polls` -- track a high-water mark and process every new scan since the last poll, oldest-first. Agent scans are skipped because they can't be reported on.
- `Scan Completion: fix 406 on report download` -- allow per-request headers and send Accept: text/csv for the report body.
- `Retry on transient connection errors` -- 3 retries with 5/15/30s backoff, only on ConnectionError / Timeout. HTTP errors aren't retried. Also fixes a TypeError where the raw exception object made it into PluginException(cause=...) and broke JSON marshalling.
- `Show numeric status code for unmapped HTTP errors` -- append the actual code to "Unknown Status Code" so things like 406 are visible in logs.
- `Bump to 8.0.17` -- version, SDK 6.4.3 → 6.5.1, changelog.

## 🧪 Testing
- Unit tests passing
- Tested live against IVM 8.0.17 -- multi-scan bursts go through, report downloads succeed, long polls survive, event reaches the orchestrator

### DIFF
--- a/plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
+++ b/plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
@@ -62,7 +62,7 @@ class ScanCompletion(insightconnect_plugin_runtime.Trigger):
             else:
                 self.logger.info(f"Found {len(new_scan_ids)} new completed scan(s) to process: {new_scan_ids}")
                 for scan_id in new_scan_ids:
-                    results = self.get_results_from_latest_scan(scan_id=int(scan_id))
+                    results = self.get_results_from_latest_scan(scan_id=scan_id)
                     self.send({Output.SCAN_ID: scan_id, Output.SCAN_COMPLETED_OUTPUT: results})
                     last_seen_scan_id = scan_id
 
--- a/plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
+++ b/plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
@@ -29,24 +29,33 @@ class ScanCompletion(insightconnect_plugin_runtime.Trigger):
         )
 
     def run(self, params={}):
-        self.logger.info("Getting latest scan...")
-
-        site_id = params.get(Input.SITE_ID)
-        resource_helper = ResourceRequests(self.connection.session, self.logger, self.connection.ssl_verify)
+        # START INPUT BINDING - DO NOT REMOVE - ANY INPUTS BELOW WILL UPDATE WITH YOUR PLUGIN SPEC AFTER REGENERATION
+        site_id = params.get(Input.SITE_ID, "")
         interval_seconds = params.get(Input.INTERVAL, 5) * 60
+        # END INPUT BINDING - DO NOT REMOVE
 
+        # Determine endpoint based on site_id
+        if site_id:
+            endpoint = endpoints.Scan.site_scans(self.connection.console_url, site_id)
+        else:
+            endpoint = endpoints.Scan.scans(self.connection.console_url)
+
+        # Initialize resource helper and last seen scan ID tracker
+        resource_helper = ResourceRequests(self.connection.session, self.logger, self.connection.ssl_verify)
         last_seen_scan_id = None
+
+        self.logger.info("Getting latest scan...")
         while True:
             # Without a high-water mark we'd paginate the entire scan history every poll, so
             # defer scanning until a baseline is established (e.g. console has no finished
             # scans yet, or only Agent scans which we skip).
             if last_seen_scan_id is None:
-                last_seen_scan_id = self.find_latest_completed_scan(site_id, resource_helper)
+                last_seen_scan_id = self.find_latest_completed_scan(endpoint, resource_helper)
                 if last_seen_scan_id is None:
                     time.sleep(interval_seconds)
                     continue
 
-            new_scan_ids = self.find_new_completed_scans(site_id, last_seen_scan_id, resource_helper)
+            new_scan_ids = self.find_new_completed_scans(endpoint, last_seen_scan_id, resource_helper)
 
             if not new_scan_ids:
                 self.logger.info("No new scans, sleeping.")
@@ -135,20 +144,15 @@ class ScanCompletion(insightconnect_plugin_runtime.Trigger):
 
         return row
 
-    def find_latest_completed_scan(self, site_id: str, resource_helper: ResourceRequests) -> Union[int, None]:
+    def find_latest_completed_scan(self, endpoint: str, resource_helper: ResourceRequests) -> Union[int, None]:
         """
-        Get the most recent reportable finished scan ID, used as the initial high-water
-        mark when the trigger starts. Two different endpoints depending on whether Site
-        ID is provided as an input or not.
+        Get the most recent reportable finished scan ID, used as the initial baseline
+        when the trigger starts.
 
-        :param site_id: Optional site id input
+        :param endpoint: The API endpoint for scans
         :param resource_helper: ResourceRequests instance for API calls
         :return: ID of the latest 'finished' scan, or None if no reportable scan exists
         """
-        if site_id:
-            endpoint = endpoints.Scan.site_scans(self.connection.console_url, site_id)
-        else:
-            endpoint = endpoints.Scan.scans(self.connection.console_url)
 
         response = resource_helper.paged_resource_request(
             endpoint=endpoint, method="get", params={"sort": "id,desc", "size": 500, "page": 0}
@@ -163,26 +167,21 @@ class ScanCompletion(insightconnect_plugin_runtime.Trigger):
         return None
 
     def find_new_completed_scans(
-        self, site_id: str, last_seen_scan_id: int, resource_helper: ResourceRequests
+        self, endpoint: str, last_seen_scan_id: int, resource_helper: ResourceRequests
     ) -> List[int]:
         """
         Return all 'finished' scan IDs greater than the last seen scan ID, ordered ascending
         so they can be processed in chronological order. Since the API returns scans sorted
-        by ID descending, pagination stops once a scan ID at or below the high-water mark
+        by ID descending, pagination stops once a scan ID at or below the last seen ID
         is encountered.
 
-        :param site_id: Optional site id input
-        :param last_seen_scan_id: High-water mark; scans with this ID or lower are skipped
+        :param endpoint: The API endpoint for scans
+        :param last_seen_scan_id: Baseline; scans with this ID or lower are skipped
         :param resource_helper: ResourceRequests instance for API calls
         :return: List of new finished scan IDs in ascending order
         """
-        if site_id:
-            endpoint = endpoints.Scan.site_scans(self.connection.console_url, site_id)
-        else:
-            endpoint = endpoints.Scan.scans(self.connection.console_url)
 
         new_scan_ids = []
-
         for page in range(self._MAX_PAGES_PER_POLL):
             response = resource_helper.resource_request(
                 endpoint=endpoint, method="get", params={"sort": "id,desc", "size": 500, "page": page}
@@ -195,11 +194,10 @@ class ScanCompletion(insightconnect_plugin_runtime.Trigger):
 
             reached_known_scan = False
             for scan in resources:
+                # Get the scan ID and if not present skip
                 scan_id = scan.get("id")
                 if scan_id is None:
                     continue
-                if self._is_reportable_finished_scan(scan):
-                    new_scan_ids.append(scan_id)
 
                 # Stop when reaching the last known scan. If last_seen_scan_id is None (first run)
                 # this will paginate through all historical scans up to the page cap
@@ -208,6 +206,10 @@ class ScanCompletion(insightconnect_plugin_runtime.Trigger):
                     reached_known_scan = True
                     self.logger.info(f"Reached known scan ID {last_seen_scan_id}, stopping pagination")
                     break
+
+                # If scan is reportable, add to list
+                if self._is_reportable_finished_scan(scan):
+                    new_scan_ids.append(scan_id)
 
             page_info = response.get("page") or {}
             total_pages = page_info.get("totalPages", 0)
--- a/plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
+++ b/plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
@@ -14,6 +14,12 @@ from komand_rapid7_insightvm.util import endpoints
 
 
 class ScanCompletion(insightconnect_plugin_runtime.Trigger):
+    # Safety bound on the per-poll pagination loop in find_new_completed_scans. At size=500
+    # this caps a single poll at 50k scans — well above any realistic burst — and prevents
+    # an infinite loop if the API keeps returning fresh pages without ever reaching the
+    # high-water mark.
+    _MAX_PAGES_PER_POLL = 100
+
     def __init__(self):
         super(self.__class__, self).__init__(
             name="scan_completion",
@@ -162,9 +168,8 @@ class ScanCompletion(insightconnect_plugin_runtime.Trigger):
             endpoint = endpoints.Scan.scans(self.connection.console_url)
 
         new_scan_ids = []
-        page = 0
 
-        while True:
+        for page in range(self._MAX_PAGES_PER_POLL):
             response = resource_helper.resource_request(
                 endpoint=endpoint, method="get", params={"sort": "id,desc", "size": 500, "page": page}
             )
@@ -188,8 +193,11 @@ class ScanCompletion(insightconnect_plugin_runtime.Trigger):
             total_pages = page_info.get("totalPages", 0)
             if reached_known_scan or (page + 1) >= total_pages:
                 break
-
-            page += 1
+        else:
+            self.logger.warning(
+                f"Reached page cap ({self._MAX_PAGES_PER_POLL}) before exhausting scan history; "
+                "some completed scans may be missed this poll"
+            )
 
         # Process oldest-first to preserve event ordering
         new_scan_ids.sort()
@@ -206,7 +214,7 @@ class ScanCompletion(insightconnect_plugin_runtime.Trigger):
         :param scan: Scan resource as returned by the InsightVM scans API
         :return: True if the scan is finished and not an Agent scan
         """
-        return scan.get("status") == "finished" and scan.get("scanType") != "Agent"
+        return scan.get("status", "").lower() == "finished" and scan.get("scanType", "").lower() != "agent"
 
 
 class ScanQueries:
--- a/plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
+++ b/plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
@@ -33,8 +33,18 @@ class ScanCompletion(insightconnect_plugin_runtime.Trigger):
 
         site_id = params.get(Input.SITE_ID)
         last_seen_scan_id = self.find_latest_completed_scan(site_id)
+        interval_seconds = params.get(Input.INTERVAL, 5) * 60
 
         while True:
+            # Without a high-water mark we'd paginate the entire scan history every poll, so
+            # defer scanning until a baseline is established (e.g. console has no finished
+            # scans yet, or only Agent scans which we skip).
+            if last_seen_scan_id is None:
+                last_seen_scan_id = self.find_latest_completed_scan(site_id)
+                if last_seen_scan_id is None:
+                    time.sleep(interval_seconds)
+                    continue
+
             new_scan_ids = self.find_new_completed_scans(site_id, last_seen_scan_id)
 
             if not new_scan_ids:
@@ -46,7 +56,7 @@ class ScanCompletion(insightconnect_plugin_runtime.Trigger):
                     self.send({Output.SCAN_ID: scan_id, Output.SCAN_COMPLETED_OUTPUT: results})
                     last_seen_scan_id = scan_id
 
-            time.sleep(params.get(Input.INTERVAL, 5) * 60)
+            time.sleep(interval_seconds)
 
     def get_results_from_latest_scan(self, scan_id: int):
         """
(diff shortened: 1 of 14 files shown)


plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
```diff
@@ -142,22 +134,79 @@ def find_latest_completed_scan(self, site_id: str, cached: bool) -> int:
...
+        if site_id:
+            endpoint = endpoints.Scan.site_scans(self.connection.console_url, site_id)
         else:
-            response = resource_helper.resource_request(endpoint=endpoint, method="get", params={"sort": "id,desc"})
+            endpoint = endpoints.Scan.scans(self.connection.console_url)
+
+        new_scan_ids = []
+        page = 0
+
+        while True:
```
igorski-r7: I'm not sure if this should be a `while True` loop or something like a `for _ in range(0, 100000)` loop to prevent an infinite loop 😃

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
```diff
@@ -142,22 +134,79 @@ def find_latest_completed_scan(self, site_id: str, cached: bool) -> int:
...
+        """
+        Determine whether a scan is finished and eligible for SQL reporting.
+        InsightVM rejects SQL reports scoped to a single scan when that scan is an
+        Insight Agent scan ("When the scope of the report is a scan, agent scans
+        may not be specified."), so those must be skipped.
+
+        :param scan: Scan resource as returned by the InsightVM scans API
+        :return: True if the scan is finished and not an Agent scan
+        """
+        return scan.get("status") == "finished" and scan.get("scanType") != "Agent"
```
igorski-r7: Imo we could lower the status as safeguard? 😃 
```suggestion
        return scan.get("status", "").lower() == "finished" and scan.get("scanType", "").lower() != "agent"
```

plugins/rapid7_insightvm/komand_rapid7_insightvm/util/resource_requests.py
```diff
@@ -73,13 +80,44 @@ def __init__(self, session: Session, logger: Logger, ssl_verify: bool) -> None:
         self.session.headers.update(self._HEADERS)
         self.ssl_verify = ssl_verify
 
+    def _send_with_retry(self, request_method, url: str, **extras):
```
igorski-r7:
```suggestion
    def _send_with_retry(self, request_method, url: str, **extras) -> None:
```

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
```diff
@@ -142,22 +134,79 @@ def find_latest_completed_scan(self, site_id: str, cached: bool) -> int:
...
-            )
+        response = resource_helper.paged_resource_request(
+            endpoint=endpoint, method="get", params={"sort": "id,desc", "size": 500, "page": 0}
+        )
 
-            for scan in response:
-                if scan.get("status") == "finished":
-                    self.logger.info(f"Latest finished scan ID: {scan.get('id')}")
-                    return scan.get("id")
+        for scan in response:
```
ekelly-r7, line 161: if there's no scan in the response, there's no return statement (so this function can return None)

type hint should be updated to reflect this at least

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
```diff
@@ -142,22 +134,79 @@ def find_latest_completed_scan(self, site_id: str, cached: bool) -> int:
...
+            resources = response.get("resources") or []
+            if not resources:
+                break
+
+            reached_known_scan = False
+            for scan in resources:
+                scan_id = scan.get("id")
+                if scan_id is None:
+                    continue
+                if last_seen_scan_id is not None and scan_id <= last_seen_scan_id:
```
ekelly-r7, line 205: if there's no `last_seen_scan_id` will this not paginate through every scan? seems very inefficient and could flood the system if there's many historical scans 

looks like this is historical behaviour, might be outside the scope of this PR to address but seems like an issue waiting to happen if someone has 1000's of scans

ekelly-r7:
Unit tests are needed for the 4 proposed fixes - especially the `find_new_completed_scans`

igorski-r7:
After updating the `Accept` headers in `Download Report` it started not to be failing. For comparison this is the response from current `master` version:

```json
"body": {
	"error": "An error occurred during plugin execution! Malformed JSON received along with a status code of Unknown Status Code Verify your connection is pointing to your local console and not. If this issue persists contact support for assistance.",
	"exception": {
		"assistance": "Verify your connection is pointing to your local console and not. If this issue persists contact support for assistance.",
		"cause": "Malformed JSON received along with a status code of Unknown Status Code" ...
```

And this is the response in current version `8.0.17`:

```json
{
"body": {
	"log": "rapid7/Rapid7 InsightVM Console:8.0.17. Step name: generate_adhoc_sql_report\nCreating report configuration with name: Rapid7-InsightConnect-AdhocReport-5ba1f443-1bf5-409c-b352-f9826486a4e3\nGenerating report\nReport ID 56587, instance ID 61900 status: complete\nDownloading SQL report contents\nCleaning up report configuration \n",
"meta": {},
"output": {
	"report": {
		"content": "YXNzZXRfaWQsbWFjX2FkZHJlc3MsaXBfYWRkcmVzcyxob3N ...
```

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
```diff
@@ -142,22 +150,84 @@ def find_latest_completed_scan(self, site_id: str, cached: bool) -> int:
...
+        so they can be processed in chronological order. Since the API returns scans sorted
+        by ID descending, pagination stops once a scan ID at or below the high-water mark
+        is encountered.
+
+        :param site_id: Optional site id input
+        :param last_seen_scan_id: High-water mark; scans with this ID or lower are skipped
+        :return: List of new finished scan IDs in ascending order
+        """
+
+        resource_helper = ResourceRequests(self.connection.session, self.logger, self.connection.ssl_verify)
```
joneill-r7: why are we creating a new resource helper here? this could be shared to the trigger?

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
joneill-r7: should this log here that we're breaking as we should have reached the correct point?

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
```diff
@@ -142,22 +150,84 @@ def find_latest_completed_scan(self, site_id: str, cached: bool) -> int:
...
+            reached_known_scan = False
+            for scan in resources:
+                scan_id = scan.get("id")
+                if scan_id is None:
+                    continue
+                if last_seen_scan_id is not None and scan_id <= last_seen_scan_id:
+                    reached_known_scan = True
+                    break
+                if self._is_reportable_finished_scan(scan):
+                    new_scan_ids.append(scan_id)
```
joneill-r7, line 212: are we missing this on the last iteration, because of the break above?

plugins/rapid7_insightvm/komand_rapid7_insightvm/util/resource_requests.py
```diff
@@ -73,13 +80,44 @@ def __init__(self, session: Session, logger: Logger, ssl_verify: bool) -> None:
...
+        transport errors only (ConnectionError, Timeout). HTTPError and other
+        RequestException subclasses are raised immediately as a PluginException.
+
+        :param request_method: Bound session method, e.g. self.session.get
+        :param url: Endpoint URL to request
+        :param extras: Keyword arguments forwarded to the session method
+        :return: requests.Response object on success
+        """
+        last_error = None
+        for attempt, backoff in enumerate((0,) + self._RETRY_BACKOFF_SECONDS):
```
joneill-r7, line 113: exactly how is this fixing the 'transient' errors? from memory when we hit this before it's how we creating the `self.session` in the connection and never closing this. this looks like once we get that error, we sleep and then try again? what's reestablishing this new session?

plugins/rapid7_insightvm/komand_rapid7_insightvm/util/util.py
```diff
@@ -76,7 +76,11 @@ def adhoc_sql_report(connection, logger, report_payload: dict):
     # Download report
     endpoint = endpoints.Report.download(connection.console_url, report_id, report_instance_id)
     logger.info("Downloading SQL report contents")
-    report_contents = resource_helper.resource_request(endpoint=endpoint, json_response=False)
+    # SQL-query report bodies are CSV; the session default of `Accept: application/json`
+    # makes IVM 8.x reject the download with HTTP 406, so override per request.
+    report_contents = resource_helper.resource_request(
+        endpoint=endpoint, json_response=False, headers={"Accept": "text/csv"}
```
joneill-r7, line 82: can we clarify this is correct? it seems to me like we used to send `application/json` and the report would be returned like this _or_ maybe it was just ignored but would return in the format the report was generated. and now we've enforced to always return in CSV format? and since the IVM API has been updated, it now returns 406 if the accept header is not supported by the report?

plugins/rapid7_insightvm/plugin.spec.yaml
```diff
@@ -6,7 +6,7 @@ title: Rapid7 InsightVM Console
 description: InsightVM is a powerful vulnerability management tool which finds, prioritizes,
   and remediates vulnerabilities. This plugin uses an orchestrator to get top remediations,
   scan results and start scans
-version: 8.0.16
+version: 8.0.17
```
joneill-r7, line 9: the 'supported versions' below should be updated? it is API v3, but should we mention the `8.48.0` version i have read during slack conversations?

plugins/rapid7_insightvm/komand_rapid7_insightvm/util/util.py
igorski-r7: I think this is fine here according to the changes that were made to the `Download Report` API 😃 By default we are sending the `Accept: application/json` headers, which in that case would cause the `Download Report` API to return `406` instead of the SQL Report. 

The SQL Report format is `csv` and according to the docs we need to override the headers to be `Accept: text/csv` to properly download it 😃

plugins/rapid7_insightvm/komand_rapid7_insightvm/util/resource_requests.py
igorski-r7: Yup, +1. Imo we should recreate the session object on retry to have it pure 😃

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
igorski-r7: Imo this could be shared to the whole trigger. I will update it 😃

plugins/rapid7_insightvm/plugin.spec.yaml
igorski-r7: Hm, wonder about including that `8.48.0`? It's not mentioned anywhere in the API public docs, so possibly it can be some internal semantic version? 🤔

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
```diff
@@ -32,20 +32,22 @@ def run(self, params={}):
...
 
+        last_seen_scan_id = self.find_latest_completed_scan(site_id, resource_helper)
+
         while True:
             # Without a high-water mark we'd paginate the entire scan history every poll, so
             # defer scanning until a baseline is established (e.g. console has no finished
             # scans yet, or only Agent scans which we skip).
             if last_seen_scan_id is None:
-                last_seen_scan_id = self.find_latest_completed_scan(site_id)
+                last_seen_scan_id = self.find_latest_completed_scan(site_id, resource_helper)
```
joneill-r7: we've got this call twice, but the while is designed to allow it to be 'None'. Can we not just delete this line on 38, and let the conditions in the loop get the latest scan?

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
```diff
@@ -171,10 +174,9 @@ def find_new_completed_scans(self, site_id: str, last_seen_scan_id: int) -> List
...
         :param site_id: Optional site id input
         :param last_seen_scan_id: High-water mark; scans with this ID or lower are skipped
+        :param resource_helper: ResourceRequests instance for API calls
         :return: List of new finished scan IDs in ascending order
         """
-
-        resource_helper = ResourceRequests(self.connection.session, self.logger, self.connection.ssl_verify)
         if site_id:
             endpoint = endpoints.Scan.site_scans(self.connection.console_url, site_id)
         else:
```
joneill-r7: this is more historical debt in this code base so not a blocking comment, but this probably could be simplified to check for site_id once before the loop and then set the endpoint and just pass this into the two methods `find_new_completed_scans` & `find_latest_completed_scan`

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
igorski-r7: We can't remove that line as `last_seen_scan_id` variable needs to be set there, but we can set it to `None` in that case 😃 

Anyway, it would require additional loop to scan it on trigger `first` run. Imo it's fine to set it to `None` and have a single place for that `find_latest_completed_scan` 😃

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
```diff
@@ -189,18 +191,24 @@ def find_new_completed_scans(self, site_id: str, last_seen_scan_id: int) -> List
...
             if not resources:
+                self.logger.info("No more resources on this page, stopping pagination")
                 break
 
             reached_known_scan = False
             for scan in resources:
                 scan_id = scan.get("id")
                 if scan_id is None:
                     continue
+                if self._is_reportable_finished_scan(scan):
```
joneill-r7: I know this is just moved based on my previous comment, but this will now include the report from the previous run and cause it to fire twice? the next condition should remove this scan ID if `scan_id <= last_seen_scan_id`?

plugins/rapid7_insightvm/komand_rapid7_insightvm/util/resource_requests.py
```diff
@@ -73,13 +80,58 @@ def __init__(self, session: Session, logger: Logger, ssl_verify: bool) -> None:
...
+        # Try to close the session
+        if isinstance(self.session, requests.Session):
+            try:
+                self.session.close()
+            except Exception as error:
+                self.logger.error(f"An exception occurred during session closing on recreation: {error}")
+
+        # Recreate the session
+        self.session = requests.Session()
+        self.session.headers.update(self._HEADERS)
```
joneill-r7, line 96: this resets the session but where are the credentials now coming from? they used to get set here? https://github.com/rapid7/insightconnect-plugins/blob/dab23078f5fc33a100862605654429f31be33046/plugins/rapid7_insightvm/komand_rapid7_insightvm/connection/connection.py#L31

plugins/rapid7_insightvm/komand_rapid7_insightvm/util/util.py
joneill-r7: Thanks - i also see in the existing logic for the report download it was expecting this to be CSV

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
igorski-r7: Imo worth including it and removing duplicates 😃

plugins/rapid7_insightvm/komand_rapid7_insightvm/util/resource_requests.py
igorski-r7: Good catch! Yup I need to restore it from the session before recreating 😄

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
igorski-r7: Hmm I think the previous order was right? Because `scan_id <= last_seen_scan_id` means it's already been processed? 😃 

So technically we need to revert it back in `if statement` ordering

plugins/rapid7_insightvm/komand_rapid7_insightvm/util/resource_requests.py
```diff
@@ -73,13 +80,66 @@ def __init__(self, session: Session, logger: Logger, ssl_verify: bool) -> None:
...
 
+    def _recreate_session(self) -> None:
+        authentication = None
+        if isinstance(self.session, requests.Session):
+            # Preserve auth before closing
+            authentication = self.session.auth
+
+            # Close the session to clear the connection pool
+            try:
+                self.session.close()
```
joneill-r7: I know session.close() still preserves this auth attribute but it makes me a bit uneasy how this is written. Is it safer to take a copy in this ever changes in future behaviour of the session lib? I could see this changing under the hood and never being caught during dev testing.

plugins/rapid7_insightvm/komand_rapid7_insightvm/util/resource_requests.py
```diff
@@ -73,13 +80,66 @@ def __init__(self, session: Session, logger: Logger, ssl_verify: bool) -> None:
...
+            except Exception as error:
+                self.logger.error(f"An exception occurred during session closing on recreation: {error}")
+
+        # Recreate the session
+        self.session = requests.Session()
+        self.session.headers.update(self._HEADERS)
+
+        # If auth was preserved, restore it
+        if authentication:
+            self.session.auth = authentication
```
joneill-r7: similar to my comment above - we're simply doing 'if authentication' and set it back onto the newly created session here. But what happens if this is not set, does the session just fail to trigger next? I think we'd be safer to have the `HTTPBasicAuth(username, password)` directly recreated? I may be being overcautious but these sessions have caused us issues in the past (and evidently in this trigger with them expiring)

plugins/rapid7_insightvm/komand_rapid7_insightvm/util/resource_requests.py
igorski-r7: Yup, imo not to overcomplicate it we can pull username, password from session object on `Resource Requests` init, and then pass it for recreation with same `Basic Auth` 😃 I will add it

plugins/rapid7_insightvm/komand_rapid7_insightvm/util/resource_requests.py
igorski-r7: I have updated the code there, and tested it out how it goes with that session recreation and it looks to be working fine I'd say 😃  This are the logs:

```
2026-06-22 07:52:46 [info     ] rapid7/Rapid7 InsightVM Console:8.0.17. Step name: list_reports
2026-06-22 07:52:47 [info     ] Trying https://instance/api/3/reports?page=0&size=10&sort=ASC
2026-06-22 07:52:47 [info     ] Forcing session recreation on first request for testing
2026-06-22 07:52:47 [info     ] Session recreated successfully with credentials
2026-06-22 07:52:47 [info     ] Trying https://instance/api/3/reports?page=0&size=10&sort=ASC
2026-06-22 07:52:48 [info     ] Trying https://instance/api/3/reports?page=1&size=10&sort=ASC
...
```

Im going to push those changes soon

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
```diff
@@ -22,33 +29,44 @@ def __init__(self):
...
+                    continue
+
+            new_scan_ids = self.find_new_completed_scans(endpoint, last_seen_scan_id, resource_helper)
+
+            if not new_scan_ids:
+                self.logger.info("No new scans, sleeping.")
+            else:
+                self.logger.info(f"Found {len(new_scan_ids)} new completed scan(s) to process: {new_scan_ids}")
+                for scan_id in new_scan_ids:
+                    results = self.get_results_from_latest_scan(scan_id=int(scan_id))
```
joneill-r7: do we need this cast to int? if it's being returned as a string then we might have a problem on `new_scan_ids.sort()`? or also in the comparison in the ID returned vs `last_seen_scan_id`? it's maybe a safeguard but maybe then needs applied across the board on these IDs

joneill-r7 (commented):
there's been a lot of logic changes outside of the 406 error handling but I don't see any new unit tests. can some be added to cover these changes?

ekelly-r7 (approved):

plugins/rapid7_insightvm/komand_rapid7_insightvm/triggers/scan_completion/trigger.py
igorski-r7: According to the InsightVM API docs that `id` comes as `int64` which in general means we don't need that additional guard and cast to `int`. Do you think its fine to remove it?

joneill-r7 (approved):
lgtm - chatted offline about following up with unit tests

PR #3973, opened 2026-06-22
Rapid7 InsightVM - Added unittests for Scan Completion trigger | Reformatted example json.resp files

igorski-r7:
## 🎫 Ticket

Ticket: SOAR-21326

## 🧩 Type of Change

- [ ] Feature
- [ ] Bug fix
- [X] Other

## 🧠 Background & Motivation

This is an update to include unit tests for the recent update to the 'Scan Completion' trigger.

## ✨ What Changed

The main focus is on improving consistency in JSON formatting (specifically, array and object indentation and structure), and adding new mock responses for error and report creation scenarios. Also, it includes additional unit tests written for `Scan Completion` trigger.

## 🧪 Testing

It was tested out locally.


plugins/rapid7_insightvm/unit_test/test_scan_completion.py
```diff
@@ -0,0 +1,284 @@
...
+
+        # Should only find reportable scans (skipping null id)
+        self.assertIn(125, scans)
+        self.assertIn(124, scans)
+
+    @patch("time.sleep")
+    @patch("requests.Session.delete", side_effect=Util.mocked_requests)
+    @patch("requests.Session.post", side_effect=Util.mocked_requests)
+    @patch("requests.Session.get", side_effect=Util.mocked_requests)
+    def test_scan_completion_skip_agent_and_unfinished_scans(
```
joneill-r7, line 139: I don't think we can start changing the trigger logic here without it getting very complicated but I could see a customer querying why a scan has been skipped. eg Scan ID 126 in the test file "scan_completion_mixed_scan_types.json.resp‎".

Within the trigger should we log, the scan ID being skipped and log the scan type / status so it's easily identified in the trigger logs? Around this logic: https://github.com/rapid7/insightconnect-plugins/commit/219148a48bb40a25344f401038d3c3d5cb9a8936#diff-dd2f8e1b6dfeae78fae742db5d534a6b17d12bd7418cbdddd6437f5cda7ac913R162

plugins/rapid7_insightvm/unit_test/util.py
```diff
@@ -154,8 +173,63 @@ def raise_for_status(self):
...
+        if kwargs.get("url") == "https://example.com/api/3/sites/6/scans":
+            return MockResponse("scan_completion_pagination_boundary", 200)
+
+        # No new scans scenario (IDs: 124, 123, 122 - all <= last_seen_scan_id)
+        if kwargs.get("url") == "https://example.com/api/3/sites/7/scans":
+            return MockResponse("scan_completion_no_new_scans", 200)
+
+        # Failure mock for exception testing: site_id=999 triggers 500 error
+        if kwargs.get("url") == "https://example.com/api/3/sites/999/scans":
+            return MockResponse("error_response", 500)
```
joneill-r7, line 211: when does this URL get passed in the new tests?

plugins/rapid7_insightvm/unit_test/util.py
```diff
@@ -154,8 +173,63 @@ def raise_for_status(self):
...
+            if not hasattr(Util, "_scan_call_count"):
+                Util._scan_call_count = 0
+            Util._scan_call_count += 1
+            if Util._scan_call_count <= 1:
+                return MockResponse("scan_completion_list_scans", 200)
+            return MockResponse("scan_completion_new_scans", 200)
+
+        # Failure mock for exception testing: site_id=999 triggers 500 error
+        if kwargs.get("url") == "https://example.com/api/3/sites/999/scans":
+            return MockResponse("error_response", 500)
```
joneill-r7, line 230: checking this is being used as well, I can't see with how these conditions are being met that we're using these mocked files and how they relate to the new tests

joneill-r7 (commented):
couple of smal questions / one suggested logger to the trigger

plugins/rapid7_insightvm/unit_test/util.py
igorski-r7: It's like a mock for `999` side ID when in input just to enforce raising an exception to catch it 😃

plugins/rapid7_insightvm/unit_test/test_scan_completion.py
igorski-r7: Yup, sure! 😃 I will add that log there then

plugins/rapid7_insightvm/unit_test/util.py
igorski-r7: Yup 😃 , its like each of these mocked urls and that `_scan_call_count` are being used in `Scan Completion` trigger.

Its not only to catch exact matches, but also to catch URLs that don't match specific routes as `Scan Completion` trigger sends requests to a couple of places. It's more like a fallback for any `/api/3/sites/*/scans` requests that don't match the specific site IDs 😃

plugins/rapid7_insightvm/unit_test/util.py
joneill-r7: I see that side ID in a test now, my bad!

joneill-r7 (approved):

PR #3987, opened 2026-06-23
Add Kerberos (SASL GSSAPI) authentication to Active Directory LDAP plugin

ewilson-r7:
## 🎫 Ticket

N/A

## 🧩 Type of Change

- [x] Feature
- [ ] Bug fix
- [ ] Other

## 🧠 Background & Motivation

Customers have raised concerns about the security of the existing NTLM-only authentication mechanism. NTLM is susceptible to relay attacks, exposes credential hashes on every connection, and lacks mutual authentication. Microsoft has been actively deprecating NTLM in favor of Kerberos. This PR adds Kerberos as an authentication option, giving customers a more secure, ticket-based auth path that aligns with Microsoft's recommended practices.

## ✨ What Changed

- Added `auth_type` connection input with options: NTLM, Kerberos, Auto (default)
- Added `kerberos` connection input for KDC address and domain name
- Plugin configures `/etc/krb5.conf`, `/etc/resolv.conf`, and acquires a TGT via `kinit` at runtime — no manual container configuration required by the customer
- Auto mode attempts Kerberos first (when config is provided), then falls back to NTLM
- Added `cryptography==48.0.0` dependency for TLS channel binding support with LDAPS
- Changed container user to root for `/etc` file write access
- Bumped version to 11.0.0 and connection_version to 10
- Fixed pre-existing test failures in enable/disable users tests (error format mismatch with current SDK)

## 🧪 Testing

- 70 unit tests passing covering Kerberos configuration, ticket acquisition, username format handling (DOMAIN\user stripped for kinit), error paths (OSError, LDAPBindError, LDAPSocketOpenError), NTLM fallback, and Auto mode behavior
- End-to-end testing against a Windows Server 2022 domain controller (TATOOINE.LAB domain):
  - Kerberos auth over LDAP (port 389) — confirmed working
  - Kerberos auth over LDAPS (port 636) — confirmed working after DC certificate setup
  - Verified Kerberos ticket acquisition via Event ID 4768/4769 on the DC
  - Tested Query, Query Group Membership, Disable User, Enable User, and Reset Password actions
Screenshot 2026-06-23 at 11 44 38 AM
Screenshot 2026-06-23 at 11 50 12 AM


plugins/active_directory_ldap/plugin.spec.yaml
```diff
@@ -30,7 +30,7 @@ hub_tags:
 sdk:
   type: slim
   version: 6.6.0
-  user: nobody
+  user: root
```
igorski-r7: Is this required? I mean, I can see that it's because of the kerberos init and config files that are written to `/etc/*`, but possibly we can write it out to temp files?

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
```diff
@@ -48,13 +53,138 @@ def __init__(
...
+        if "\\" in username:
+            username = username.split("\\", 1)[1]
+
+        kinit_command = f"echo '{self.password}' | kinit {username}@{upper_domain}"
+        with subprocess.Popen(
+            kinit_command,
+            shell=True,  # noqa: B602
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        ) as process:
```
igorski-r7: Well I think this could be handled differently. The `shell=True` with the password interpolated into the string is a shell injection vuln as of `echo '{password}'` puts the plaintext password in the process args (ps, /proc/<pid>/cmdline).

What about using the `gssapi` and its `acquire_cred_with_password` instead of that kinit command etc? Something like

```python
import gssapi
from gssapi.raw import acquire_cred_with_password

name = gssapi.Name(f"{username}@{realm}", gssapi.NameType.user)
creds = acquire_cred_with_password(name, self.password.encode("utf-8")).creds
creds.store(usage="initiate", overwrite=True)  # into KRB5CCNAME
```

If this is something that we need to leave kinit command and running it in subprocess I think its worth note to use no shell with no stdin password like

```python
with subprocess.Popen(
    ["kinit", f"{username}@{upper_domain}"],
    stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
) as process:
    stdout, stderr = proc.communicate(input=(self.password + "\n").encode())
```

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
```diff
@@ -48,13 +53,138 @@ def __init__(
...
+            stdout, stderr = process.communicate()
+
+        if process.returncode != 0:
+            self.logger.error(f"kinit failed - stdout: {stdout.decode('utf-8')}, stderr: {stderr.decode('utf-8')}")
```
igorski-r7: Please don't log the entire `stdeer` as in that case due to `echo` password can get leaked into the logger

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
```diff
@@ -48,13 +53,138 @@ def __init__(
...
+        """Write configuration content to a system file."""
+        try:
+            with open(filepath, "w", encoding="utf-8") as config_file:
+                config_file.write(content)
+        except OSError as error:
+            raise PluginException(
+                cause=f"Failed to write {description}.",
+                assistance=f"Ensure the plugin container has write access to {filepath}.",
+                data=error,
+            ) from error
```
igorski-r7: Writing to /etc/krb5.conf requires root and mutates shared container state. I suggest to write to a temp file and point Kerberos at it via env var.

```python
krb5 = tempfile.NamedTemporaryFile(mode="w", suffix=".conf", delete=False, encoding="utf-8")
krb5.write(krb5_config); krb5.flush()
os.environ["KRB5_CONFIG"] = krb5.name
```

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
```diff
@@ -48,13 +53,138 @@ def __init__(
...
+            f"\n"
+            f"[domain_realm]\n"
+            f".{domain} = {upper_domain}\n"
+            f"{domain} = {upper_domain}\n"
+        )
+        self._write_config_file("/etc/krb5.conf", krb5_config, "Kerberos configuration")
+
+    def _write_network_config(self, kdc: str, domain: str):
+        """Write DNS and hosts configuration for Kerberos network resolution."""
+        self._write_config_file("/etc/resolv.conf", f"search {domain}\nnameserver {kdc}\n", "DNS configuration")
```
igorski-r7: I would say the same with temp files instead (even temp files using context manager)

plugins/active_directory_ldap/unit_test/test_manage_users_common.py
igorski-r7, line 1: Is this required? Why it got moved out of the previous unittests?

plugins/active_directory_ldap/requirements.txt
```diff
@@ -4,5 +4,6 @@
 pycryptodome==3.20.0
 parameterized==0.9.0
 gssapi==1.9.0
+cryptography==48.0.0
```
igorski-r7, line 7: Why do we need it here?

plugins/active_directory_ldap/unit_test/test_kerberos_connection.py
igorski-r7, line 1: Nice to see that many unittests here, but wonder if its possible to write them more short and readable? Also, please include typehints for test methods.

plugins/active_directory_ldap/unit_test/test_enable_users.py
```diff
@@ -6,48 +6,11 @@
...
-            (
-                {Input.DISTINGUISHED_NAMES: ["CN=Users,DC=example,DC=com"]},
-                {
-                    Output.COMPLETED: ["CN=Users,DC=example,DC=com"],
-                    Output.FAILED: [],
-                },
-            ),
-        ]
-    )
+    @parameterized.expand(MANAGE_USERS_TEST_CASES)
```
igorski-r7, line 13: Why it was moved to separate const variable?

igorski-r7 (changes requested):
A couple of comments left 😃 

The core problem is architectural. I mean, kerberos is being set up by shelling out to kinit command and writing to root system files (`/etc/krb5.conf`, `/etc/resolv.conf`, `/etc/hosts`). This requires user to be updated in the container to root.

This chain produces couple of issues. Running as root, shell injection via `shell=True`, plaintext password exposure in the process `args/logs`, and writing out to shared system files. Because every action is wrapped in `@with_connection` decorator, this setup also reruns on every action, so those `/etc/hosts` grows and concurrent actions clobber each other's config.

plugins/active_directory_ldap/requirements.txt
ewilson-r7: Required by ldap3's SASL/GSSAPI implementation to compute TLS channel bindings when connecting over LDAPS. Without it, Kerberos+LDAPS fails with `LDAPPackageUnavailableError: package cryptography missing`. Added an inline comment explaining this in the updated commit.

plugins/active_directory_ldap/unit_test/test_manage_users_common.py
ewilson-r7: SonarQube flagged >3% duplication between `test_disable_users.py` and `test_enable_users.py` — they had identical parameterized test data. Extracting to a shared module eliminates the duplication while keeping both test files functional.

plugins/active_directory_ldap/unit_test/test_enable_users.py
ewilson-r7: Same reason as above — SonarQube duplication gate. Both enable/disable test files had identical test case data, so it's referenced from a single shared constant.

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
ewilson-r7: Good call — completely reworked. Now using `gssapi.raw.acquire_cred_with_password` directly (pure Python). No subprocess, no shell, no password in process args or logs. The `subprocess` import is removed entirely.

plugins/active_directory_ldap/plugin.spec.yaml
ewilson-r7: No longer required — switched to writing krb5.conf to a temp file (`tempfile.NamedTemporaryFile`) and setting `KRB5_CONFIG` env var. Container now runs as `nobody` again. No `/etc` writes at all.

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
ewilson-r7: Done — now uses `tempfile.NamedTemporaryFile(delete=False)` with `os.environ['KRB5_CONFIG']` pointing to it. No root needed.

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
ewilson-r7: Removed entirely — `/etc/resolv.conf` and `/etc/hosts` writes are gone. The KDC is contacted by IP from the connection input, so no DNS mutation is needed.

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
ewilson-r7: Fixed — no subprocess output is logged anymore. The gssapi approach raises a `GSSError` exception which is caught and re-raised as a PluginException with only the error type (no credentials in the message).

plugins/active_directory_ldap/unit_test/test_kerberos_connection.py
ewilson-r7: Tightened up — added type hints to all test methods, extracted a `mock_temp_file()` helper, and organized into focused test classes (Configuration, DomainResolution, EstablishConnection, ErrorHandling).

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
```diff
@@ -48,14 +57,117 @@ def __init__(
...
         set_library_log_detail_level(ERROR)
         set_library_log_hide_sensitive_data(True)
 
+    @staticmethod
+    def _is_ip_address(host: str) -> bool:
+        try:
+            ipaddress.ip_address(host)
+            return True
+        except ValueError:
+            return False
```
igorski-r7, line 74: To do so you can use `validators` package. But not required to update is fine.

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
```diff
@@ -48,14 +57,117 @@ def __init__(
...
+            f"{upper_domain} = {{\n"
+            f"kdc = {kdc}\n"
+            f"admin_server = {kdc}\n"
+            f"default_domain = {upper_domain}\n"
+            f"}}\n"
+            f"\n"
+            f"[domain_realm]\n"
+            f".{domain} = {upper_domain}\n"
+            f"{domain} = {upper_domain}\n"
+        )
```
igorski-r7: Couldn't this be moved to some `constants.py` or something with a template to create that file? Like with multiline str that can be used to format the structure properly via `.format(...)`

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
```diff
@@ -48,14 +57,117 @@ def __init__(
...
+        with tempfile.NamedTemporaryFile(mode="w", suffix=".conf", delete=False, encoding="utf-8") as krb5_file:
```
igorski-r7: That encoding `utf-8` could be added to constants as well 😃

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
ewilson-r7: Done — moved to `util/constants.py` as `KRB5_CONFIG_TEMPLATE` using `.format(realm=..., kdc=..., domain=...)` placeholders.

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
ewilson-r7: Added `ENCODING = "utf-8"` to the same `constants.py` and referenced it in both the tempfile write and the gssapi password encoding.

plugins/active_directory_ldap/komand_active_directory_ldap/util/api.py
ewilson-r7: Sticking with `ipaddress` from stdlib for now since it's zero-dependency and handles both IPv4/IPv6. Happy to switch to `validators` in a follow-up if we adopt it more broadly.

igorski-r7 (approved):

PR #4016, opened 2026-07-16
Python 3 Script - Action `Run`: Added per connection dependency isolation | Fixed issue related to `ModuleNotFound` when plugin was updated | Updated SDK to the latest version (6.6.0)

igorski-r7:
## 🎫 Ticket

Ticket: SOAR-21666

## 🧩 Type of Change

- [ ] Feature
- [X] Bug fix
- [X] Other

## 🧠 Background & Motivation

Users encountered `ModuleNotFoundError` when updating the plugin, making it unreliable for running Python scripts with third-party dependencies. This update fixes dependency resolution and adds per-connection isolation to prevent module conflicts between connections.

## ✨ What Changed

- Fixed ModuleNotFoundError: Improved dependency detection and installation so third-party modules are reliably available when scripts run.
- Per-connection virtual environments: Each connection now gets its own isolated virtual environment, preventing module conflicts when different connections use different dependencies.
- Credential security: User credentials are now passed via environment variables instead of command-line arguments, removing exposure to system monitoring tools.
- Injection protection: Invalid module specifiers starting with - are now rejected to prevent pip argument injection.

## 🧪 Testing

It was tested out locally so far manually + with load tests using locust. It was working fine with separation per connection. I haven't noticed any functional break.

### DIFF
--- a/plugins/python_3_script/icon_python_3_script/connection/connection.py
+++ b/plugins/python_3_script/icon_python_3_script/connection/connection.py
@@ -101,7 +101,11 @@ class Connection(insightconnect_plugin_runtime.Connection):
             # The Path.rename is atomic on POSIX within the same filesystem
             # If another worker already placed a complete environment just reuse it
             if not final_directory.exists():
-                temporary_directory.rename(final_directory)
+                try:
+                    temporary_directory.rename(final_directory)
+                except OSError:
+                    if not final_directory.exists():
+                        raise RuntimeError(f"Failed to finalize the Python environment at {final_directory}.")
 
     def install_dependencies(self) -> None:
         try:
--- a/plugins/python_3_script/icon_python_3_script/connection/connection.py
+++ b/plugins/python_3_script/icon_python_3_script/connection/connection.py
@@ -1,9 +1,10 @@
 import sys
 import os
+import importlib.util
 from subprocess import CalledProcessError, TimeoutExpired, check_output, run  # noqa: B404
 
 import insightconnect_plugin_runtime
-from insightconnect_plugin_runtime.exceptions import ConnectionTestException
+from insightconnect_plugin_runtime.exceptions import ConnectionTestException, PluginException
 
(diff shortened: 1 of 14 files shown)


lcwiklinski-r7 (approved):

plugins/python_3_script/icon_python_3_script/connection/connection.py
```diff
@@ -1,64 +1,127 @@
...
+                raise RuntimeError("Installing Python dependencies exceeded timeout. Consider increasing timeout.")
+            except CalledProcessError as error:
+                stderr = error.stderr.decode(DEFAULT_ENCODING) if error.stderr else ""
+                raise RuntimeError(f"Non-zero exit code returned. Message: {stderr}")
+
+            # Atomically promote to the final location only when fully installed
+            # The Path.rename is atomic on POSIX within the same filesystem
+            # If another worker already placed a complete environment just reuse it
+            if not final_directory.exists():
+                temporary_directory.rename(final_directory)
```
ekelly-r7: Edge case there is a TOCTOU window between `exists()` and `rename()`. Verified on POSIX: Path.rename onto a non-empty existing directory raises `OSError: [Errno 66] Directory not empty` (it does not silently overwrite). If two workers finish nearly simultaneously, worker A renames successfully and worker B's `rename()` raises `OSError`.

This `OSError` is not caught (`_run_pip_install` catches only `TimeoutExpired` and `CalledProcessError`; `install_dependencies` and `ensure_dependencies` catch only `RuntimeError`. In `run()` (action.py:41-42), `ensure_dependencies()` is called outside the try/except at line 48, so a rename-race `OSError` surfaces as an unhandled exception instead of `PluginException`.

igorski-r7: Updated in the latest commit. Thanks a lot! Good catch 😃

ekelly-r7 (approved):

PR #4025, opened 2026-07-27
Rapid7 InsightIDR: retry transient 5xx responses with backoff

ekelly-r7:
## 🎫 Ticket
[SOAR-21651: Intermittent 500 Errors with the InsightIDR Plugin](https://rapid7.atlassian.net/browse/SOAR-21651)

## 🧩 Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Other

## 🧠 Background & Motivation
Customers intermittently received `500 Internal Server Error` responses from the InsightIDR plugin, where re-running the same job typically succeeded on the next attempt. A newly created record can take a short time to become searchable, so a plugin follow-up request can arrive during that short indexing window and get a transient 5xx. The agreed workaround is to make the plugin resilient to these transient errors with a short delay plus retry-with-backoff.

## ✨ What Changed
- Added a centralized retry-with-backoff around the shared send path used by both `resource_request()` and `make_request()` in `util/resource_helper.py`, so every affected action benefits from a single change rather than per-action patches.
- Retries on transient 5xx (500, 502, 503, 504), bounded to 3 attempts with backoff over a ~5s window to cover the indexing lag. 4xx continues to fail fast, and after retries are exhausted the original `PluginException` is raised so failures are still surfaced.
- Retries are scoped to requests that are safe to repeat — idempotent methods plus read-only search POSTs — so non-idempotent writes (e.g. create investigation/comment, upload attachment) are never duplicated by a retry.
- Correlation ID and existing error logging are preserved across attempts; each retry is logged.
- Bumped plugin version \`12.0.8\` → \`12.0.9\` and added a \`version_history\` entry. Already on the latest SDK (\`6.6.0\`), so no SDK bump. Regenerated plugin artifacts via \`insight-plugin refresh\`.

## 🧪 Testing
- New unit tests in \`unit_test/test_resource_helper.py\` cover: 5xx-then-200 succeeds after retry; persistent 5xx raises after N attempts; 4xx not retried; non-idempotent create POST not retried. \`time.sleep\` is mocked so tests stay fast.
- Full unit suite passes on the SDK-matched Python 3.13.13 (119 passed).
- \`insight-plugin refresh\` regenerated cleanly (only expected generated files changed).
- Container smoke test: image builds and gunicorn boots cleanly in HTTP mode (Listening / Booting worker, no traceback).
- [ ] Staging integration tests (run when the release branch deploys)

### DIFF
--- a/plugins/rapid7_insightidr/komand_rapid7_insightidr/util/constants.py
+++ b/plugins/rapid7_insightidr/komand_rapid7_insightidr/util/constants.py
@@ -16,12 +16,15 @@ DEFAULT_ERROR_MESSAGE = "Unknown error occurred. Please contact support or try a
 # gaps) resolves the vast majority of these.
 RETRY_MAX_ATTEMPTS = 3
 RETRY_BACKOFF_SECONDS = [2, 3]
+# Additive jitter (0 to N seconds) on top of each backoff so concurrent jobs that hit a
+# 5xx at the same moment don't retry in lockstep. Kept additive (never subtractive) so a
+# retry can't fire before the ~1-2s indexing window the backoff is sized to cover.
+RETRY_JITTER_SECONDS = 1
 RETRYABLE_STATUS_CODES = {500, 502, 503, 504}
-# Only retry requests that are safe to repeat. Idempotent methods can be re-sent
-# without side effects; POST is excluded (a committed-then-5xx create must not be
-# duplicated) EXCEPT for read-only search endpoints, which are the primary source
-# of the transient 5xx (e.g. investigations/_search after a just-created record).
-IDEMPOTENT_METHODS = {"GET", "HEAD", "OPTIONS", "PUT", "DELETE"}
+# Only retry requests that are safe to repeat. POST is excluded (a committed-then-5xx
+# create must not be duplicated) EXCEPT for read-only search endpoints, which are the
+# primary source of the transient 5xx (e.g. investigations/_search after a just-created
+# record). All other methods the plugin issues (GET, DELETE) are idempotent.
 RETRYABLE_POST_ENDPOINT_SUFFIXES = ("_search", "/search")
 
 # Constants for Get New Alerts trigger.py
--- a/plugins/rapid7_insightidr/komand_rapid7_insightidr/util/constants.py
+++ b/plugins/rapid7_insightidr/komand_rapid7_insightidr/util/constants.py
@@ -20,6 +20,8 @@ RETRY_BACKOFF_SECONDS = [2, 3]
 # 5xx at the same moment don't retry in lockstep. Kept additive (never subtractive) so a
 # retry can't fire before the ~1-2s indexing window the backoff is sized to cover.
 RETRY_JITTER_SECONDS = 1
+# Transient/gateway 5xx worth retrying. Permanent 5xx (e.g. 501, 505) are excluded since
+# retrying cannot help.
 RETRYABLE_STATUS_CODES = {500, 502, 503, 504}
 # Only retry requests that are safe to repeat. POST is excluded (a committed-then-5xx
 # create must not be duplicated) EXCEPT for read-only search endpoints, which are the
--- a/plugins/rapid7_insightidr/komand_rapid7_insightidr/util/constants.py
+++ b/plugins/rapid7_insightidr/komand_rapid7_insightidr/util/constants.py
@@ -10,5 +10,19 @@ THREE_MONTHS_SECONDS = 7776000
 # Constants for resource_helper.py
 DEFAULT_ERROR_MESSAGE = "Unknown error occurred. Please contact support or try again later."
 
+# Retry/backoff for transient 5xx responses from InsightIDR. A newly created record can
+# take a short time to become searchable (~1-2s indexing lag), so a follow-up request can
+# transiently return a 5xx. Retrying with backoff over a ~5s window (2s + 3s across the
+# gaps) resolves the vast majority of these.
+RETRY_MAX_ATTEMPTS = 3
+RETRY_BACKOFF_SECONDS = [2, 3]
+RETRYABLE_STATUS_CODES = {500, 502, 503, 504}
+# Only retry requests that are safe to repeat. Idempotent methods can be re-sent
+# without side effects; POST is excluded (a committed-then-5xx create must not be
+# duplicated) EXCEPT for read-only search endpoints, which are the primary source
+# of the transient 5xx (e.g. investigations/_search after a just-created record).
+IDEMPOTENT_METHODS = {"GET", "HEAD", "OPTIONS", "PUT", "DELETE"}
+RETRYABLE_POST_ENDPOINT_SUFFIXES = ("_search", "/search")
+
 # Constants for Get New Alerts trigger.py
 TOTAL_SIZE = 100
--- a/plugins/rapid7_insightidr/komand_rapid7_insightidr/util/resource_helper.py
+++ b/plugins/rapid7_insightidr/komand_rapid7_insightidr/util/resource_helper.py
@@ -1,6 +1,7 @@
 import asyncio
 import base64
 import json
+import random
 import re
 import time
 import uuid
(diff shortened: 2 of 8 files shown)


plugins/rapid7_insightidr/komand_rapid7_insightidr/util/constants.py
```diff
@@ -10,5 +10,19 @@
...
+# transiently return a 5xx. Retrying with backoff over a ~5s window (2s + 3s across the
+# gaps) resolves the vast majority of these.
+RETRY_MAX_ATTEMPTS = 3
+RETRY_BACKOFF_SECONDS = [2, 3]
+RETRYABLE_STATUS_CODES = {500, 502, 503, 504}
+# Only retry requests that are safe to repeat. Idempotent methods can be re-sent
+# without side effects; POST is excluded (a committed-then-5xx create must not be
+# duplicated) EXCEPT for read-only search endpoints, which are the primary source
+# of the transient 5xx (e.g. investigations/_search after a just-created record).
+IDEMPOTENT_METHODS = {"GET", "HEAD", "OPTIONS", "PUT", "DELETE"}
```
igorski-r7: Do we need to point out those methods?

plugins/rapid7_insightidr/komand_rapid7_insightidr/util/resource_helper.py
```diff
@@ -100,6 +108,63 @@ def __init__(self, connection_headers, logger):
...
+        :param endpoint: The request URL (path is inspected for search endpoints)
+        :return: True if the request may be retried
+        """
+        method = method.upper()
+        if method in IDEMPOTENT_METHODS:
+            return True
+        if method == "POST":
+            path = endpoint.split("?", 1)[0]
+            return path.endswith(RETRYABLE_POST_ENDPOINT_SUFFIXES)
+        return False
```
igorski-r7: Just wondering if we need that method?

plugins/rapid7_insightidr/komand_rapid7_insightidr/util/resource_helper.py
```diff
@@ -100,6 +108,63 @@ def __init__(self, connection_headers, logger):
...
+        retryable = self._is_retryable_request(request.method, request.url)
+        with requests.Session() as session:
+            prepared_request = session.prepare_request(request)
+            response = session.send(prepared_request)
+            if not retryable:
+                return response
+            for attempt in range(1, RETRY_MAX_ATTEMPTS):
+                if response.status_code not in RETRYABLE_STATUS_CODES:
+                    return response
+                backoff = RETRY_BACKOFF_SECONDS[min(attempt - 1, len(RETRY_BACKOFF_SECONDS) - 1)]
```
lmacoun-r7, line 156: we should include jitter to the backoff

plugins/rapid7_insightidr/komand_rapid7_insightidr/util/constants.py
```diff
@@ -10,5 +10,19 @@
 # Constants for resource_helper.py
 DEFAULT_ERROR_MESSAGE = "Unknown error occurred. Please contact support or try again later."
 
+# Retry/backoff for transient 5xx responses from InsightIDR. A newly created record can
+# take a short time to become searchable (~1-2s indexing lag), so a follow-up request can
+# transiently return a 5xx. Retrying with backoff over a ~5s window (2s + 3s across the
+# gaps) resolves the vast majority of these.
+RETRY_MAX_ATTEMPTS = 3
+RETRY_BACKOFF_SECONDS = [2, 3]
+RETRYABLE_STATUS_CODES = {500, 502, 503, 504}
```
lmacoun-r7, line 25: missing 429, and is there is reason why not to cover all 5xx? something like 

if status_code == 429 || status_code >= 500

plugins/rapid7_insightidr/komand_rapid7_insightidr/util/resource_helper.py
ekelly-r7: The rationale is that the intermittent 500s the customer saw were on search calls, so that's the ones we want to retry.

This method was created to deliberately not retry write POSTs (could cause an issue with duplicate records)

Thoughts on why we don't need it?

plugins/rapid7_insightidr/komand_rapid7_insightidr/util/constants.py
ekelly-r7: I can simplify this so that we only guard on POST, no need for this constant

plugins/rapid7_insightidr/komand_rapid7_insightidr/util/resource_helper.py
ekelly-r7: @igorski-r7 i simplified this function a bit, but I believe we still need it. 

The guard exists to block write POSTs (create/comment/upload) from retrying, so a create that commits server-side before returning a `5xx` isn't duplicated. Read-only search `POSTs` (`_search`) are the ones we do let retry, since they carry no write side effect and are the source of the intermittent 500s.

plugins/rapid7_insightidr/unit_test/test_resource_helper.py
```diff
@@ -0,0 +1,109 @@
...
+    return response
+
+
+class TestResourceHelperRetry(TestCase):
+    def setUp(self) -> None:
+        self.helper = ResourceHelper({}, logging.getLogger("test"))
+        self.endpoint = "https://us.api.insight.rapid7.com/idr/v2/investigations/_search"
+
+    @patch("time.sleep", return_value=None)
+    @patch("requests.Session.send")
```
joneill-r7: minor but I'd patch these once at the start to avoid having to do it each test

plugins/rapid7_insightidr/komand_rapid7_insightidr/util/constants.py
ekelly-r7: Good thought on 429 - I decided to hold off on adding it here, though. This fix is scoped to the intermittent 500s we've seen on search calls, and 429s haven't been an issue for this plugin from what I can tell. I also don't believe this retry would handle a real rate limit gracefully (backoff is a fixed ~5s window - designed for the search delay). If 429's become a problem, the right tool would be the SDK rate_limiting decorator. 

On "cover all 5xx" I kept the enumerated set {500, 502, 503, 504} rather than >= 500, since 501/505 are permanent failures where a retry can't help. Updated docstring to mention this.

plugins/rapid7_insightidr/komand_rapid7_insightidr/util/resource_helper.py
```diff
@@ -100,6 +109,62 @@ def __init__(self, connection_headers, logger):
...
+        read-only search POSTs (e.g. investigations/_search), which are the primary
+        source of the transient indexing-lag 5xx and carry no write side effect. Every
+        other method the plugin issues (GET, DELETE) is idempotent and safe to re-send.
+
+        :param method: HTTP method for the request
+        :param endpoint: The request URL (path is inspected for search endpoints)
+        :return: True if the request may be retried
+        """
+        if method.upper() == "POST":
+            path = endpoint.split("?", 1)[0]
```
joneill-r7: is this split needed? I don't know what way these URLs are being formatted but don't see in the unit tests or on the DNR. It reads more like we'd want this on a GET request if we weren't retrying all GET anyway

ekelly-r7: GET will be retried - only POST writes will be guarded against 

I guess this could be simplied to just check if the path contains `RETRYABLE_POST_ENDPOINT_SUFFIXES`

joneill-r7: yeah I'm following that, but I'm just not sure where in the plugin we build `?` in the endpoints file https://github.com/rapid7/insightconnect-plugins/blob/a9e77eb00aca0557b72ed13424423c08cf2027bb/plugins/rapid7_insightidr/komand_rapid7_insightidr/util/endpoints.py to need this split in the first place

ekelly-r7: good call! should be resolved now

joneill-r7 (approved):

igorski-r7 (approved):

PR #4033, opened 2026-08-10
[SOAR-21803] Fix MISP Search for Tag trigger failing

landrisek-r7:
## 🎫 Ticket
Ticket: [SOAR-21803 — The MISP 6.0 plugin trigger is failing](https://rapid7.atlassian.net/browse/SOAR-21803) (customer report: [SI-35006](https://rapid7.atlassian.net/browse/SI-35006))

## 🧩 Type of Change
- [ ] Feature
- [x] Bug fix
- [ ] Other

## 🧠 Background & Motivation
The `Search for Tag` trigger fails on every run with:

```
TypeError: search_index() got an unexpected keyword argument 'tag'
```

Two defects in `komand_misp/triggers/search_for_tag/trigger.py`:

1. It called `client.search_index(tag=tag)`, but PyMISP `2.4.194` (the pinned/supported version) exposes the parameter as **`tags`** (plural). The trigger's call therefore raised a `TypeError` before any events were fetched, so jobs were never kicked off.
2. It then unwrapped `events["response"]`, but `search_index()` returns a **bare list** of event dicts — there is no `response` wrapper, so this would fail even after fixing (1).

The `Search Events` action already calls `search_index(..., tags=...)` correctly and iterates the returned list directly, which is why the plugin's actions worked while the trigger did not. This change brings the trigger in line with that working pattern.

## ✨ What Changed
- Fixed the trigger to call `client.search_index(tags=tag)` and iterate the returned list directly (removed the invalid `events["response"]` unwrap).
- Added unit tests for the trigger (`unit_test/test_search_for_tag.py`).
- Bumped plugin version `6.0.0 → 6.0.1` and added a Version History entry.

## 🧪 Testing
**Root cause verified against source**
- Confirmed against the PyMISP `v2.4.194` source that `search_index()` takes `tags` (not `tag`) and returns a bare `list` of event dicts (no `response` key). This matches the existing `search_events` action.

**Unit tests (added in this PR)**
Ran locally in an isolated virtualenv with the plugin's pinned deps (`insightconnect-plugin-runtime`, `jsonschema`, `pymisp==2.4.194`):

```
$ python -m unittest test_search_for_tag -v
test_search_for_tag_calls_search_index_with_tags ... ok
test_search_for_tag_event_missing_id_raises ... ok
test_search_for_tag_no_events_does_not_send ... ok
----------------------------------------------------------------------
Ran 3 tests in 0.004s
OK
```

Cases covered:
- **Happy path / regression guard** — asserts the client is called with `tags=` (plural) and that returned event IDs are sent. This is the direct guard for this bug.
- **Empty result** — no events returned ⇒ nothing is sent.
- **Error path** — an event missing its `id` surfaces a `KeyError` (logged, not silently dropped).

**Regression guard proven effective** — with the fix reverted (`tags=` → `tag=`), the tests fail as expected:
```
AssertionError: expected call not found.
Expected: search_index(tags='test-tag')
  Actual: search_index(tag='test-tag')
```

**Validation**
- `insight-plugin validate` passes `VersionValidator`, `HelpInputOutputValidator`, and checksum validation. The version bump was done via `insight-plugin refresh` and the change set trimmed to the minimal fix.
- Note: the pre-existing `RuntimeValidator` warning (base image `6.1.0` is not the latest SDK) is intentionally **out of scope** for this focused bug fix and pre-dates this branch.

[SI-35006]: https://rapid7.atlassian.net/browse/SI-35006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/misp/komand_misp/triggers/search_for_tag/trigger.py
```diff
@@ -22,12 +22,7 @@ def run(self, params={}):  # noqa: MC0001
 
             client = self.connection.client
             event_id = []
-            events = client.search_index(tag=tag)
-            try:
-                events = events["response"]
-            except KeyError:
-                self.logger.error("Unexpected search return, %s")
-                raise
+            events = client.search_index(tags=tag)
```
joneill-r7, line 25: do we not need to access the key `response` anymore?

joneill-r7 (commented):
has this been tested against a live API and not only in unit tests?

plugins/misp/komand_misp/triggers/search_for_tag/trigger.py
landrisek-r7: Correct — no more `["response"]`. In the pinned pymisp (2.4.194) `search_index` returns a flat `list[dict]` of event metadata directly (it returns `normalized_response` from `events/index`), not a `{"response": [...]}` envelope — so we iterate the return value directly. The old `events["response"]` would actually have raised `TypeError` on that list, not `KeyError`.

The core fix here is the kwarg: the parameter is `tags`, not `tag` (see the `search_index` signature), so the previous `search_index(tag=tag)` silently dropped the tag filter, which is why the trigger was failing.

joneill-r7: Has this been tested at all? I'd actually be surprised if this trigger starts up successfully as I think there's a missing ownership change in the Dockerfile to allow the trigger to read a cached file at start up.

landrisek-r7:
Good call — yes, this revision was validated locally against a live MISP instance, not just the unit tests.

I seeded an event with a test tag and ran the trigger the way the runtime does: the plugin image running as the non-root user, with the state-cache volume mounted, driven by a real `trigger_start` message. It connected, executed `Search for Tag`, and emitted the matching event id.

On the core fix: against the live API, the previous `search_index` call reproduced the customer-reported failure, and the corrected call returns the event list directly (this PyMISP version returns a plain list rather than a wrapped `response` object), which is what the trigger now relies on.

On the startup point you raised — you're right. The runtime initializes trigger state at startup under the cache directory, which the non-root user couldn't write to, so it failed before `run()`. This revision grants that user access to the cache directory, and with that in place the trigger starts cleanly and persists state as expected. CI (PluginValidator + UnitTest) is green on the current head.

igorski-r7 (approved):

lcwiklinski-r7 (changes requested):
whenever you run `insight-plugin refresh` it will change the `Dockerfile`, once it's fixed it lgtm

plugins/misp/komand_misp/triggers/search_for_tag/trigger.py
landrisek-r7: Good catch — you were right on both. Tested against a live MISP (not just unit tests): running the trigger as the non-root user with the state-cache volume mounted, it initially failed at startup because the runtime writes trigger state under `/workspace/cache`, which the non-root user couldn't write to. Adding the cache-directory ownership fixes it — the trigger then starts, runs `Search for Tag`, and emits the matching event id, with state persisted correctly.

Rather than hand-editing the Dockerfile (which `insight-plugin refresh` would overwrite), this is now driven by `enable_cache: true` in the spec, so the chown is regenerated on every refresh.

landrisek-r7:
Fixed — instead of patching the `Dockerfile` directly, I added `enable_cache: true` to the spec. `insight-plugin refresh` now regenerates the `/workspace/cache` ownership line from the template, so it survives future refreshes. Verified by stripping the line and re-running `refresh` (it comes back), and the Dockerfile output is unchanged from the previous revision.

joneill-r7 (approved):

lcwiklinski-r7 (approved):

PR #4041, opened 2026-08-24
Sophos Central - Fix issue where could fail to find agents beyond the first page of results | SDK bump to 6.6.0

igorski-r7:
## 🎫 Ticket
SOAR-21878

## 🧩 Type of Change
- [x] Bug fix
- [ ] Feature
- [ ] Other

## 🧠 Background & Motivation
Sophos Central pagination logic used `index > total_pages` as loop-exit condition combined with `page_key is None`. Comparing a page index against a total-pages count doesn't reliably terminate pagination and could stop before all pages were fetched, causing agents on pages beyond the first to be missed by `get_agent_details` and endpoint search.

Also `get_endpoints` params were being overwritten (`params = {...}`) instead of merged when both `lastSeenAfter` and page key were needed, and the page key param name (`pageKey`) didn't match the API's expected `pageFromKey`.

## ✨ What Changed
- Simplified pagination termination to `if not page_key: break` in both `util/api.py` and `actions/get_agent_details/action.py`, removing the unreliable `total_pages`/index comparison.
- Fixed `params` dict in `get_endpoints` to update (not overwrite) when adding `lastSeenAfter`, and corrected the page-key param name to `pageFromKey`.
- Bumped SDK to 6.6.0 (Dockerfile, setup.py, .CHECKSUM, plugin.spec.yaml, help.md regenerated).

## 🧪 Testing
- Added unit test `test_get_agent_details.py` with new fixtures (`get_agent_details_page_1.json.resp`, `get_agent_details_page_2.json.resp`) covering multi-page pagination for `get_agent_details`, including a not-found case.
- Existing unit test suite passes.


plugins/sophos_central/unit_test/util.py
```diff
@@ -70,12 +70,18 @@ def mock_request(*args, **kwargs):
                     return MockResponse(200, "get_blocked_items.json.resp")
                 if params == {"pageTotal": False}:
                     return MockResponse(200, "get_blocked_items_no_params.json.resp")
+                if params == {"page": 1, "pageSize": 100, "pageTotal": True}:
+                    return MockResponse(200, "get_blacklists_page_1.json.resp")
```
lcwiklinski-r7, line 74: These three mock routes point at fixtures that don't exist in `responses/` (`get_blacklists_page_1/2.json.resp` here and on line 76, `unblacklist.json.resp` on line 84), and nothing currently exercises them. It looks like `test_blacklist.py` and its fixtures were meant to land with this PR but got left out.

plugins/sophos_central/icon_sophos_central/actions/blacklist/action.py
```diff
@@ -16,37 +16,36 @@ def __init__(self):
...
-                        uuid = e.get("id")
-                        break
-
-                if uuid is None:
-                    raise PluginException(
-                        cause="Unable to unblacklist a hash that is not in the blacklist.",
-                        assistance="Please provide a hash that is already blacklisted.",
-                    )
+            uuid = None
+            for blocked_item in self.connection.client.iter_blacklist_items():
```
lcwiklinski-r7, line 37: Iterating `iter_blacklist_items()` across all pages is the main fix in this PR, but it ships with no test — there's no `test_blacklist.py`, and `iter_blacklist_items` / `_paginate_by_page_number` are uncovered too. Could we add cases for a multi-page unblacklist match and the not-in-blacklist `NOT_FOUND` path?

plugins/sophos_central/icon_sophos_central/util/api.py
```diff
@@ -12,34 +52,54 @@ def __init__(self, url, client_id, client_secret, tenant_id, version, logger):
...
-        return endpoint_id
+    def get_endpoint_id(self, entity):
+        for page in self._paginate_by_key(lambda page_key: self.get_endpoints(page_key=page_key)):
+            for item in page.get("items", []):
+                if matches_entity(entity, item):
+                    return item.get("id")
+
+        raise PluginException(preset=PluginException.Preset.NOT_FOUND)
+
+    def get_all_endpoints(self, since: str = None) -> List[dict]:
```
lcwiklinski-r7, line 89: `get_all_endpoints` / `get_all_alerts` are new and untested — the only pagination path with coverage is `get_endpoint_id` via Get Agent Details. A small multi-page aggregation test would help pin down the new all-pages behavior.

plugins/sophos_central/icon_sophos_central/actions/get_endpoints/action.py
```diff
@@ -14,4 +14,8 @@ def __init__(self):
         )
 
     def run(self, params={}):
-        return {Output.ITEMS: self.connection.client.get_endpoints(params.get(Input.SINCE)).get("items", [])}
+        # START INPUT BINDING - DO NOT REMOVE - ANY INPUTS BELOW WILL UPDATE WITH YOUR PLUGIN SPEC AFTER REGENERATION
+        since = params.get(Input.SINCE)
+        # END INPUT BINDING - DO NOT REMOVE
+
+        return {Output.ITEMS: self.connection.client.get_all_endpoints(since)}
```
lcwiklinski-r7, line 21: This switches Get Endpoints from returning just the first page to returning every page via `get_all_endpoints` — a real output change for existing workflows and potentially a large response on big tenants. Was that intended? If so, it's worth calling out in the version history alongside the Get Agent Details fix.

plugins/sophos_central/plugin.spec.yaml
```diff
@@ -50,6 +50,7 @@ hub_tags:
   keywords: [sophos, firewall]
   features: []
 version_history:
+  - "4.4.2 - Fix issue where `Get Agent Details` could fail to find agents beyond the first page of results | SDK bump to 6.6.0"
```
lcwiklinski-r7: This 4.4.2 entry only mentions Get Agent Details, but the PR also makes Get Endpoints return all pages and rewrites the Blacklist action's pagination. Worth expanding the entry so the changelog reflects those user-facing changes.

plugins/sophos_central/help.md
```diff
@@ -532,7 +375,7 @@ Example input:
 
 ```
 {
-  "from_date": "2019-09-23T12:02:01.700Z"
+  "from_date": "2019-09-23 12:02:01.700000+00:00"
```
lcwiklinski-r7: This example (and `since` on line 583) now renders as a Python datetime repr instead of ISO-8601, which contradicts the field's own description (`2019-09-23T12:02:01.700Z`) and the spec examples. A value copy-pasted in this format may be rejected by the Sophos API — worth regenerating so help.md matches the spec.

plugins/sophos_central/help.md
```diff
@@ -1,6 +1,6 @@
 # Description
 
-[Sophos Central](https://www.sophos.com) is a unified console for managing Sophos products. Using the Sophos Central plugin for Rapid7 InsightConnect, users can get alerts, endpoints, and SIEM events
+Sophos Central is a unified console for managing Sophos products. Using the Sophos Central plugin for Rapid7 InsightConnect, users can get alerts, endpoints, and SIEM events
```
lcwiklinski-r7: The `[Sophos Central](https://www.sophos.com)` link in the description dropped to plain text. It just tracks the spec (which has no link), so it's a regeneration side effect — if the link was intended, add it back in `plugin.spec.yaml` and regenerate.

plugins/sophos_central/icon_sophos_central/util/api.py
```diff
@@ -12,34 +52,54 @@ def __init__(self, url, client_id, client_secret, tenant_id, version, logger):
...
+        return [
+            item
+            for page in self._paginate_by_key(lambda page_key: self.get_endpoints(since=since, page_key=page_key))
+            for item in page.get("items", [])
+        ]
+
+    def get_all_alerts(self, since: str = None) -> List[dict]:
```
lcwiklinski-r7, line 92: `get_all_endpoints` and `get_all_alerts` are identical apart from the fetch lambda. A tiny helper like `_collect_items(fetch_page)` returning the flattened items would keep the flattening in one place — optional.

plugins/sophos_central/unit_test/test_get_agent_details.py
```diff
@@ -0,0 +1,47 @@
...
+from insightconnect_plugin_runtime.exceptions import PluginException
+
+
+@patch("requests.request", side_effect=Util.mock_request)
+class TestGetAgentDetails(TestCase):
+    @classmethod
+    def setUpClass(cls) -> None:
+        cls.action = Util.default_connector(GetAgentDetails())
+
+    @parameterized.expand(
```
lcwiklinski-r7, line 23: `@parameterized.expand` here wraps a single case, so the parameterization is mostly overhead — a plain `test_get_agent_details(self, mock_request)` reading the input/expected files inline would be simpler. (Minor: `test_name` still feeds the subtest id, so it's not entirely unused.)

lcwiklinski-r7 (commented):
Automated review pass (correctness, test coverage, docs, plus a simplify/deslop lens). Nice fixes in here — correcting `pageKey` → `pageFromKey`, no longer overwriting the params dict (which dropped `pageTotal`), the blacklist no longer raising `NOT_FOUND` after page 1, and the `if search else None` guard (the field is `required: false`).

Main thing worth resolving before merge: the rewritten blacklist pagination — the core change here — ships with no test, and `util.py` references three fixture files that aren't in the repo.

_Optional non-blocking nit (pre-existing, so not inline):_ in `get_agent_details/action.py` the "multiple agents found" log concatenates two adjacent f-strings with no space (`…first match` glues onto the previous sentence) and the second literal doesn't need the `f` prefix.

plugins/sophos_central/setup.py
```diff
@@ -1,14 +1,14 @@
...
+    name="sophos_central-rapid7-plugin",
+    version="4.4.2",
+    description="Sophos Central is a unified console for managing Sophos products. Using the Sophos Central plugin for Rapid7 InsightConnect, users can get alerts, endpoints, and SIEM events",
+    author="rapid7",
+    author_email="",
+    url="",
+    packages=find_packages(),
+    install_requires=["insightconnect-plugin-runtime"],  # Add third-party dependencies to requirements.txt, not here!
+    entry_points={"console_scripts": ["icon_sophos_central = bin.icon_sophos_central:main"]},
+)
```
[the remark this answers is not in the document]
igorski-r7: `setup.py` is generated, header literally says `# GENERATED BY INSIGHT-PLUGIN - DO NOT EDIT`. The `entry_points` block came straight out of `insight-plugin refresh` on SDK 6.6.0, same as every other plugin already on that SDK (`active_directory_ldap`, `carbon_black_cloud`, etc). The runtime invokes `bin/icon_sophos_central` directly via the Dockerfile entrypoint, not through the console script, so nothing regresses here. If the template itself is wrong that is an `insight-plugin` fix, out of scope for this bug.

plugins/sophos_central/icon_sophos_central/actions/get_endpoint_groups/action.py
```diff
@@ -16,20 +16,32 @@ def __init__(self):
...
-            "endpointIds": params.get(Input.ENDPOINTIDS),
+            "sort": sort,
+            "fields": fields,
+            "page": page,
+            "pageTotal": page_total,
+            "pageSize": page_size,
+            "ids": ids,
+            "search": "+".join(search.split(" ")),
+            "searchFields": search_fields,
+            "endpointIds": endpoint_ids,
```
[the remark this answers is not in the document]
igorski-r7: Good catch, already handled in the current head. It now reads `"+".join(search.split(" ")) if search else None` and `clean()` drops it when unset. Comment was against an earlier push.

plugins/sophos_central/help.md
```diff
@@ -382,9 +323,7 @@ Example output:
 
 ```
 {
-  "tamper_status": {
-    "enabled": true
-  }
+  "tamper_status": {}
 }
```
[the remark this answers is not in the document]
igorski-r7: Already updated, help.md now shows `{"tamper_status": {"enabled": true}}`. Stale against the latest push.

plugins/sophos_central/unit_test/util.py
```diff
@@ -70,12 +70,18 @@ def mock_request(*args, **kwargs):
...
+                if params == {"page": 2, "pageSize": 100, "pageTotal": True}:
+                    return MockResponse(200, "get_blacklists_page_2.json.resp")
```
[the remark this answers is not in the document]
igorski-r7: You are right, `test_blacklist.py` and its three fixtures got left behind when I split the commit. Adding `get_blacklists_page_1.json.resp`, `get_blacklists_page_2.json.resp` and `unblacklist.json.resp` along with the test in the next push.

plugins/sophos_central/unit_test/util.py
igorski-r7: Correct, `test_blacklist.py` and those three fixtures got left behind when I split the commit. Adding them in the next push.

plugins/sophos_central/icon_sophos_central/actions/blacklist/action.py
igorski-r7: Agreed, that is the gap. Adding `test_blacklist.py` with a multi page unblacklist where the hash only appears on page 2, the not in blacklist `NOT_FOUND` path, the invalid hash path, and a blacklist happy path. That also covers `iter_blacklist_items` and `_paginate_by_page_number`.

plugins/sophos_central/icon_sophos_central/util/api.py
igorski-r7: Fair. Adding multi page aggregation tests for both so the flattening and the cursor advance are pinned, not just the `get_endpoint_id` path.

plugins/sophos_central/icon_sophos_central/actions/get_endpoints/action.py
igorski-r7: Intended. Before this PR `Get Endpoints` returned only page 1, and `Get Alerts` looped on `has_more`, a key the Sophos response never contains, so it also stopped at page 1. Both silently truncated on any tenant bigger than one page, which is the same class of bug SI-35109 reported for `Get Agent Details`. Output schema is unchanged, only the item count grows. Calling it out in the version history as you suggest.

plugins/sophos_central/plugin.spec.yaml
igorski-r7: Will expand the 4.4.2 entry to cover `Get Endpoints` and `Get Alerts` returning all pages and the `Blacklist` paging rewrite, not just `Get Agent Details`.

plugins/sophos_central/help.md
```diff
@@ -532,7 +373,7 @@ Example input:
 
 ```
 {
-  "from_date": "2019-09-23T12:02:01.700Z"
+  "from_date": "2019-09-23 12:02:01.700000+00:00"
 }
```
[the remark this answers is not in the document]
igorski-r7: Real regression and it starts in the spec. `example: 2019-09-23T12:02:01.700Z` is unquoted in `plugin.spec.yaml`, so YAML parses it as a timestamp and the generator renders the Python repr into help.md. Quoting both examples and regenerating puts ISO 8601 back in the docs.

plugins/sophos_central/help.md
igorski-r7: Root cause is in the spec. `example: 2019-09-23T12:02:01.700Z` is unquoted, so YAML parses it as a timestamp and the generator writes the Python repr into help.md. Quoting both examples (`since` and `from_date`) and regenerating restores ISO 8601.

plugins/sophos_central/help.md
igorski-r7: Side effect of regeneration, the spec `description` has no link and help.md now just mirrors it. The link was intended, so I am putting `[Sophos Central](https://www.sophos.com)` back at the front of the spec description and regenerating.

plugins/sophos_central/icon_sophos_central/util/api.py
igorski-r7: Sure, folding the shared flattening into one small helper so `get_all_endpoints` and `get_all_alerts` only differ by the fetch lambda.

plugins/sophos_central/unit_test/test_get_agent_details.py
igorski-r7: Will keep `parameterized.expand` but with real cases in it, the not found case moves into the same table, which is where it earns its keep.

lcwiklinski-r7 (approved):

PR #4048, opened 2026-08-27
Release Automox 3.1.1

jstarynec-r7:
## 🎫 Ticket

SI-35110 / SOAR-21928

## 🧩 Type of Change

- [x] Bug fix
- [x] Chore (SDK bump — required by the build gate, see below)

## 🧠 Background & Motivation

`upload_vulnerability_sync_file` passed **both** a `preset` and a custom `assistance` to `PluginException`. The SDK overwrites `assistance` whenever a `preset` is set (`insightconnect_plugin_runtime/exceptions.py`), so the response code and body were computed and then discarded. Every rejected upload rendered the same canned text no matter what Automox returned:

```
An error occurred during plugin execution! Server error occurred Verify your plugin connection
inputs are correct and not malformed and try again. If the issue persists, please contact support.
```

Verified against both SDK 6.5.1 and 6.6.0 that the old code reproduces that string character-for-character. This action is also the only Automox action that bypasses `_call_api`, so it never logged the status code either — leaving nothing in the job log or InsightOps.

**This has now cost two investigations on the same symptom:**

- **SOAR-18350** — *"insightConnect: Automox 'Upload Vulnerability Sync File' returning 'server error'"*, opened 2024-12-04, closed **Abandoned** with no diagnosis.
- **SI-35110** — opened 2026-08, three weeks of back-and-forth. The canned "malformed" wording was read as Automox's own response, sending support after CSV column requirements, with no way to see the actual status code. A 400 was also being reported to the customer as a "server error".

The pairing is rare and looks unintentional: repo-wide there are **12** occurrences of `preset=` together with `assistance=`, and **three of those twelve are in this one function**. The working idiom (`preset=` with `data=`, which *is* preserved) appears **575** times.

## ✨ What Changed

**The fix (`fix(automox): report the Automox response...`)**

- **Response code and body now survive.** `cause`/`assistance`/`data` are passed explicitly instead of alongside a `preset`. Applied to all three error branches (non-2xx, `JSONDecodeError`, generic).
- **Request and status are logged**, mirroring `_call_api` — so the status code is recoverable from logs even if the step error is truncated.
- **A 4xx is no longer labelled a server error.**
- **Fixed a latent `UnboundLocalError`**: the `JSONDecodeError` and generic handlers both read `response.status_code`, but `response` is unbound when `requests.post` itself fails (connection error/timeout). That path would have raised `UnboundLocalError` instead of a usable message.
- **Payload shape logging** in the action: decoded size, delimiter presence, ASCII — **never content**. An already-base64 input decodes without error and uploads a base64 blob instead of a CSV; this distinguishes the two without putting customer data in logs. Bounded slice, so a 1M-row report isn't copied just to log it.
- **Tests**: added a 400 case and assertions that the status code and body actually reach the caller. The previous error-path test asserted only `cause`, which equals the preset cause by construction — which is why this shipped green. Also pinned the SDK's preset-overwrites-assistance behaviour so the workaround can't be silently undone.

> **Design note — why not the 575-instance `preset=` + `data=` idiom?** It preserves the body but keeps the preset's *cause*, so a 400 would still read "Server error occurred", and the status code would still be absent. Passing `cause`/`assistance`/`data` explicitly surfaces both and matches `_call_api` in this same file (`api_client.py:62-65`), which is the local precedent. Happy to switch if reviewers prefer repo-wide consistency over the extra detail.

**The SDK bump (`chore(automox): bump SDK to 6.6.0...`)** — required to land the fix, not opportunistic cleanup:

- `RuntimeValidator` rejects any base image that isn't the latest tag. 6.6.0 shipped after 3.1.0 released on 6.5.1, so the build gate fails for **any** new Automox PR until this happens. The first push of this PR failed CI on exactly that.
- `CloudReadyValidator` separately requires `enable_cache: true` on a `cloud_ready` plugin declaring a trigger. This plugin has `get_events` and had been shipping `enable_cache: false`.
- Kept as a separate commit so the behavioural fix can be reviewed independently of the mechanical bump.

## 🧪 Testing

- **115 unit tests pass** against SDK 6.6.0 (and previously against 6.5.1).
- `black --check --line-length 120` clean on all changed files (black 26.3.1, matching CI).
- `insight-plugin validate` (1.11.0) passes everything except `DockerValidator`, which fails only because the Docker daemon was unavailable locally — CI builds the image and passes.
- Confirmed against the installed SDK that a mocked 400 now renders `Failed to upload file to Vulnerability Sync. Automox returned response code 400. Response was <body>`, where it previously rendered only the canned preset text.
- **Re-verified the premise under 6.6.0**: the SDK still overwrites a custom `assistance` when a `preset` is supplied, so the workaround is still required.
- Assertions are on `assistance`/`data` individually rather than the whole message — an early iteration passed a message-wide substring check only because a `TypeError` was masked by the generic handler.

**Not tested:** no Automox tenant or API key available, so this is not exercised against the live API.

**Out of scope, noted for follow-up:** the remaining 9 occurrences of the same anti-pattern, in `get_url` (3), `ping` (5) and `microsoft_teams` (1).

This change does not attempt to fix the customer's upload — it makes the cause visible. Root cause on SI-35110 should follow from the next run on 3.1.1.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


plugins/automox/icon_automox/actions/upload_vulnerability_sync_file/action.py
```diff
@@ -38,6 +38,16 @@ def run(self, params={}):
                 assistance=f"Review uploaded CSV and ensure it is a base64 encoded input. Error: {str(error)}",
             )
 
+        # Shape only, never content: an already-base64 input decodes without error but uploads a
+        # base64 blob instead of a CSV, which Automox rejects with no clue as to why. A blob is
+        # ASCII with no delimiter, where a real header row has one. Bounded slice so a 1M-row
+        # report is not copied just to log it.
+        prefix = decoded_data[:200]
+        self.logger.info(
+            f"Decoded CSV: {len(decoded_data)} bytes, delimiter in first {len(prefix)} bytes: "
```
lmacoun-r7: is this log customer facing?

lmacoun-r7 (dismissed):

landrisek-r7 (approved):

PR #4049, opened 2026-08-27
SOAR-21874: `get_new_investigations` allow for API indexing latency

joneill-r7:
## 🎫 Ticket

[SOAR-21874](https://rapid7.atlassian.net/browse/SOAR-21874)

## 🧩 Type of Change

- [ ] Feature
- [X] Bug fix
- [ ] Other

## 🧠 Background & Motivation

Customer reported that since `12.0.2` there was a fault in our `Get New Investigations` trigger was not finding investigations on the IDR API. When looking at the trigger logger pre-12.0.1 you can see that the trigger always did a 20 minute look back and deduped the returning investigations. In 12.0.2 this approach was changed assuming that we could use a normal time sliding window of `time-a -> time-b` followed by `time-b -> time-c` but this does not account for the latency on the IDR API.

## ✨ What Changed

- Updated that we now overlap with a 5 minute window to allow for delayed indexing on IDR API. 20 minutes seems excessive and in other PRs (#4027) we were told this indexing should only be a matter of seconds. 
- No unit tests existed at all - I have added some basic tests to cover this new dedupe logic + time querying so that we would catch this if it's ever changed in future. 

Post initial commit from code review comments:
- State now contains a last_poll_time to correctly resume after the container restarts 
- Extended the code coverage in the unit tests to cover more dedupe / lack of in more test scenarios
- Unit tests now patch _save_state() to allow running unit tests in debug mode - mock on time.sleep() was throwing exceptions in debug mode via IDE debug mode. 

## 🧪 Testing

- Unit tests
- Manual testing; more logs available on jira ticket:
  - restarting the docker container; reads from the state file and uses this cache to filter any new investigations:
```
Connect: Connecting...
Loaded state from /workspace/cache/example-trigger-id-johnny-n.json: {'RRNs': ['rrn:investigation:us:cf946362-0809-426d-987c-849bde704c16:investigation:JFWFO2ATG70X', 'rrn:investigation:us:cf946362-0809-426d-987c-849bde704c16:investigation:OIQXB1BROI5I']}
rapid7/Rapid7 InsightIDR:12.0.10. Step name: get_new_investigations
Get Investigations: trigger started
Investigations search criteria: []
Initial poll time set to: '2026-08-31T08:26:14.972726+00:00'
Searching for new investigations from '2026-08-31T08:26:14.972726+00:00' to '2026-08-31T08:36:09.973879+00:00'
Making request to https://us.api.insight.rapid7.com/idr/v2/investigations/_search with request ID: 05220345-4b62-4c04-a1f9-278d5567dcef
No new investigations found.
Checkpoint for next iteration 2026-08-31T08:31:09.973879+00:00 to allow for API latency
Sleeping for 15 seconds...
```
  - retrieving investigation + **with the 5 minute look back on next run from now** (post failure expected as no local env running to get data)
 ```
Sleeping for 15 seconds...
Searching for new investigations from '2026-08-31T08:34:52.857891+00:00' to '2026-08-31T08:40:08.817652+00:00'
Making request to https://us.api.insight.rapid7.com/idr/v2/investigations/_search with request ID: 90330996-9163-436d-8bea-c60af0aeb5bf
Retrieved total of 1 investigations.
Duplicate investigation found and skipped: rrn:investigation:us:cf946362-0809-426d-987c-849bde704c16:investigation:RFL38H4K1J7W
Saving investigations RRNs retrieved during this execution.
Checkpoint for next iteration 2026-08-31T08:35:08.817652+00:00 to allow for API latency
Sleeping for 15 seconds...

# next run now returns TWO investigations because of the overlap retrunning the same one above:
Searching for new investigations from '2026-08-31T08:35:08.817652+00:00' to '2026-08-31T08:40:24.671715+00:00'
Making request to https://us.api.insight.rapid7.com/idr/v2/investigations/_search with request ID: e61b8235-535e-4959-bece-27e84001d941
Retrieved total of 2 investigations.
```

  - subsequent runs and deduping already retrieved investigation
 ```
Saving investigations RRNs retrieved during this execution.
Checkpoint for next iteration 2026-08-31T08:35:24.671715+00:00 to allow for API latency
Sleeping for 15 seconds...
Searching for new investigations from '2026-08-31T08:35:24.671715+00:00' to '2026-08-31T08:40:40.749397+00:00'
Making request to https://us.api.insight.rapid7.com/idr/v2/investigations/_search with request ID: 7b1f242a-de42-45fc-ba81-97985ede210e
Retrieved total of 2 investigations.
Duplicate investigation found and skipped: rrn:investigation:us:cf946362-0809-426d-987c-849bde704c16:investigation:88NYN2QE6WPY
Duplicate investigation found and skipped: rrn:investigation:us:cf946362-0809-426d-987c-849bde704c16:investigation:RFL38H4K1J7W
Saving investigations RRNs retrieved during this execution.
Checkpoint for next iteration 2026-08-31T08:35:40.749397+00:00 to allow for API latency
Sleeping for 15 seconds...
```

  - container restart holds the checkpoint:
```
No new investigations found.
Checkpoint for next iteration 2026-08-31T14:41:33.417123+00:00 to allow for API latency
Sleeping for 15 seconds...

# forced restart below

[2026-08-31 14:46:41 +0000] [1] [INFO] Handling signal: term
[2026-08-31 14:46:51 +0000] [1] [INFO] Starting gunicorn 25.0.0
[2026-08-31 14:46:51 +0000] [1] [INFO] Listening at: http://0.0.0.0:10001 (1)
[2026-08-31 14:46:51 +0000] [1] [INFO] Using worker: gthread
[2026-08-31 14:46:51 +0000] [8] [INFO] Booting worker with pid: 8
Connect: Connecting...
Loaded state from /workspace/cache/example-trigger-id-johnny-n.json: {'last_poll_time': '2026-08-31T14:41:33.417123+00:00', 'RRNs': []}
rapid7/Rapid7 InsightIDR:12.0.10. Step name: get_new_investigations
Get Investigations: trigger started
Investigations search criteria: []
Detected a container restart, resumming from last poll time: 2026-08-31T14:41:33.417123+00:00
Searching for new investigations from '2026-08-31T14:41:33.417123+00:00' to '2026-08-31T14:47:10.610603+00:00'
```

[SOAR-21874]: https://rapid7.atlassian.net/browse/SOAR-21874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/plugins/rapid7_insightidr/help.md
+++ b/plugins/rapid7_insightidr/help.md
@@ -3428,6 +3428,7 @@ Example output:
 
 # Version History
 
+* 12.0.10 - Trigger `Get New Investigations` updated for IDR API latency issues and dedupe logic added | Updated depdancies
 * 12.0.9 - Retry requests with backoff on transient InsightIDR 5xx responses to resolve intermittent 500 errors
 * 12.0.8 - Enabled cache to configure the appropriate permissions in the Dockerfile
 * 12.0.7 - Updated dependencies | SDK bump to 6.6.0
--- a/plugins/rapid7_insightidr/help.md
+++ b/plugins/rapid7_insightidr/help.md
@@ -3428,7 +3428,7 @@ Example output:
 
 # Version History
 
-* 12.0.10 - Trigger `Get New Investigations` updated for IDR API latency issues and dedupe logic added | Updated depdancies
+* 12.0.10 - Trigger `Get New Investigations` updated for IDR API latency issues and dedupe logic added | Updated dependencies
 * 12.0.9 - Retry requests with backoff on transient InsightIDR 5xx responses to resolve intermittent 500 errors
 * 12.0.8 - Enabled cache to configure the appropriate permissions in the Dockerfile
 * 12.0.7 - Updated dependencies | SDK bump to 6.6.0
--- a/plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
+++ b/plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
@@ -12,8 +12,12 @@ from datetime import datetime, UTC, timedelta
 from typing import Dict, Any, List
 
 DEFAULT_FREQUENCY_SECONDS = 15
-INITIAL_LOOKBACK_MINUTES = 1
+INITIAL_LOOKBACK_MINUTES = 10
+# set this to be half of the INITIAL_LOOKBACK_MINUTES so the second run doesn't expand the window beyond lookback
+API_LATENCY_OVERLAP_MINUTES = 5
 MAX_NUMBER_OF_RETRIES = 20
+
+STATE_KEY = "RRNs"
 
 
 class GetNewInvestigations(insightconnect_plugin_runtime.Trigger):
@@ -26,22 +30,27 @@ class GetNewInvestigations(insightconnect_plugin_runtime.Trigger):
             output=GetNewInvestigationsOutput(),
         )
 
+    @staticmethod
+    def get_current_time() -> datetime:
+        """Returns current UTC time - extracted for test mocking without adding freezegun dependency to the plugin"""
+        return datetime.now(UTC)
+
     def run(self, params={}):
         # START INPUT BINDING - DO NOT REMOVE - ANY INPUTS BELOW WILL UPDATE WITH YOUR PLUGIN SPEC AFTER REGENERATION
         search = params.get(Input.SEARCH)
         frequency = params.get(Input.FREQUENCY, DEFAULT_FREQUENCY_SECONDS)
         # END INPUT BINDING - DO NOT REMOVE
 
-        # Initialize the trigger starting point
-        retry_attempts_counter = 0
-        last_poll_time = datetime.now(UTC) - timedelta(minutes=INITIAL_LOOKBACK_MINUTES)
+        # Initialize the trigger starting point & set to dedupe investigations RRNs
+        retry_attempts_counter, prev_investigations = 0, set(self.state.get(STATE_KEY, []))
+        last_poll_time = self.get_current_time() - timedelta(minutes=INITIAL_LOOKBACK_MINUTES)
         self.logger.info("Get Investigations: trigger started")
         self.logger.info(f"Investigations search criteria: {search}")
         self.logger.info(f"Initial poll time set to: '{last_poll_time.isoformat()}'")
 
         while True:
-            # Calculate current time for this iteration (delay by 5 second to avoid overlap) and log search details
-            current_time = datetime.now(UTC) - timedelta(seconds=5)
+            # Calculate current time for this iteration (delay 5s to ensure time is safely in past before API indexes)
+            current_time = self.get_current_time() - timedelta(seconds=5)
             self.logger.info(
                 f"Searching for new investigations from '{last_poll_time.isoformat()}' to '{current_time.isoformat()}'"
             )
@@ -68,17 +77,20 @@ class GetNewInvestigations(insightconnect_plugin_runtime.Trigger):
             # If investigations were found, log total and send them one by one
             # Otherwise, log that no new investigations were found
             if investigations:
-                self.logger.info(f"Found total of {len(investigations)} new investigations.")
-                for investigation in investigations:
-                    self.send_investigation(investigation)
+                self.logger.info(f"Retrieved total of {len(investigations)} investigations.")
+                prev_investigations = self.dedupe_and_send_investigations(investigations, prev_investigations)
             else:
                 self.logger.info("No new investigations found.")
 
             # Update last poll time and reset retry counter for next iteration
-            # Overlap by 1 second to handle IDR API's open time boundaries (exclusive start_time/end_time)
-            # This ensures investigations created at exact boundary timestamps are included in the next poll
-            last_poll_time = current_time
+            # Overlap window by 5 min to catch late-indexed investigations (IDR indexing is often slow)
+            last_poll_time = current_time - timedelta(minutes=API_LATENCY_OVERLAP_MINUTES)
+            self.logger.info(f"Checkpoint for next iteration {last_poll_time.isoformat()} to allow for API latency")
             retry_attempts_counter = 0
+
+            # save the state for fallback in case of plugin restart
+            self.state[STATE_KEY] = list(prev_investigations)
+            self._save_state()
 
             # Back off before next iteration
             self.logger.info(f"Sleeping for {frequency} seconds...")
@@ -119,8 +131,26 @@ class GetNewInvestigations(insightconnect_plugin_runtime.Trigger):
                 investigations.extend(response.get("data", []))
         return investigations
 
-    def send_investigation(self, investigation: Dict[str, Any]) -> None:
-        self.logger.info(f"Investigation found: {investigation.get('rrn', 'N/A')}")
+    def dedupe_and_send_investigations(self, investigations: List[Dict[str, Any]], previous_investigations: set) -> set:
+        latest_investigations = set()
+        for investigation in investigations:
+            rrn = investigation.get("rrn", "N/A")
+            # Track all RRNs from current poll for state management (enables deduplication across restarts)
+            latest_investigations.add(rrn)
+            if rrn not in previous_investigations:
+                self.send_investigation(investigation, rrn)
+            else:
+                self.logger.info(f"Duplicate investigation found and skipped: {rrn}")
+        # make sure we don't return an empty set if no new investigations were found, so we don't lose the previous state
+        if latest_investigations:
+            self.logger.info("Saving investigations RRNs retrieved during this execution.")
+            previous_investigations = latest_investigations
+        else:
+            self.logger.info("No new investigations were sent. Retaining existing state for next iteration.")
+        return previous_investigations
+
+    def send_investigation(self, investigation: Dict[str, Any], rrn: str) -> None:
+        self.logger.info(f"Investigation found: {rrn}")
         self.send({Output.INVESTIGATION: clean(investigation)})
 
     @staticmethod
--- a/plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
+++ b/plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
@@ -18,6 +18,7 @@ API_LATENCY_OVERLAP_MINUTES = 5
 MAX_NUMBER_OF_RETRIES = 20
 
 STATE_KEY = "RRNs"
+TIME_STATE_KEY = "last_poll_time"
 
 
 class GetNewInvestigations(insightconnect_plugin_runtime.Trigger):
@@ -43,10 +44,17 @@ class GetNewInvestigations(insightconnect_plugin_runtime.Trigger):
 
         # Initialize the trigger starting point & set to dedupe investigations RRNs
         retry_attempts_counter, prev_investigations = 0, set(self.state.get(STATE_KEY, []))
-        last_poll_time = self.get_current_time() - timedelta(minutes=INITIAL_LOOKBACK_MINUTES)
+
         self.logger.info("Get Investigations: trigger started")
         self.logger.info(f"Investigations search criteria: {search}")
-        self.logger.info(f"Initial poll time set to: '{last_poll_time.isoformat()}'")
+
+        if last_poll_time := self.state.get(TIME_STATE_KEY):
+            self.logger.info(
+                f"Detected a container restart, resumming from last poll time: {last_poll_time.isoformat()}"
+            )
+        else:
+            last_poll_time = self.get_current_time() - timedelta(minutes=INITIAL_LOOKBACK_MINUTES)
+            self.logger.info(f"Initial poll time set to: '{last_poll_time.isoformat()}'")
 
         while True:
             # Calculate current time for this iteration (delay 5s to ensure time is safely in past before API indexes)
@@ -89,6 +97,7 @@ class GetNewInvestigations(insightconnect_plugin_runtime.Trigger):
             retry_attempts_counter = 0
 
             # save the state for fallback in case of plugin restart
+            self.state[TIME_STATE_KEY] = last_poll_time
             self.state[STATE_KEY] = list(prev_investigations)
             self._save_state()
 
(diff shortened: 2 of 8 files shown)


lcwiklinski-r7 (commented):
Assisted code review (correctness + simplification/deslop + tests + docs). Main thing to look at is the restart-deduplication question (state horizon vs initial lookback); the rest is a dead branch, a couple of misleading comments, a user-facing typo, and some test-coverage gaps. Left inline — nothing blocking beyond the restart-dedup question.

plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
```diff
@@ -12,9 +12,13 @@
 from typing import Dict, Any, List
 
 DEFAULT_FREQUENCY_SECONDS = 15
-INITIAL_LOOKBACK_MINUTES = 1
+INITIAL_LOOKBACK_MINUTES = 10
+# set this to be half of the INITIAL_LOOKBACK_MINUTES so the second run doesn't expand the window beyond lookback
```
lcwiklinski-r7, line 16: This rationale doesn't quite hold — steady-state windows are `frequency + API_LATENCY_OVERLAP_MINUTES` wide and never reset to the lookback, so "half of the lookback" isn't what bounds the window. Could you restate it as "5 min of tolerance for late-indexed investigations"?

plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
```diff
@@ -26,22 +30,27 @@ def __init__(self):
...
         search = params.get(Input.SEARCH)
         frequency = params.get(Input.FREQUENCY, DEFAULT_FREQUENCY_SECONDS)
         # END INPUT BINDING - DO NOT REMOVE
 
-        # Initialize the trigger starting point
-        retry_attempts_counter = 0
-        last_poll_time = datetime.now(UTC) - timedelta(minutes=INITIAL_LOOKBACK_MINUTES)
+        # Initialize the trigger starting point & set to dedupe investigations RRNs
+        retry_attempts_counter, prev_investigations = 0, set(self.state.get(STATE_KEY, []))
+        last_poll_time = self.get_current_time() - timedelta(minutes=INITIAL_LOOKBACK_MINUTES)
```
lcwiklinski-r7: On restart this looks back `INITIAL_LOOKBACK_MINUTES` (10 min), but the restored state only holds the previous poll's RRNs (~5 min of window), so investigations 5–10 min old get re-sent as duplicates on the first post-restart poll. Should the lookback be capped at the overlap horizon so the restored state actually covers what gets re-fetched?

plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
```diff
@@ -119,8 +131,26 @@ def get_investigations(
                 investigations.extend(response.get("data", []))
         return investigations
 
-    def send_investigation(self, investigation: Dict[str, Any]) -> None:
-        self.logger.info(f"Investigation found: {investigation.get('rrn', 'N/A')}")
+    def dedupe_and_send_investigations(self, investigations: List[Dict[str, Any]], previous_investigations: set) -> set:
+        latest_investigations = set()
+        for investigation in investigations:
+            rrn = investigation.get("rrn", "N/A")
```
lcwiklinski-r7: Using `"N/A"` as the fallback dedup key means any two investigations that both lack an `rrn` collapse to the same key, so the second gets skipped as a duplicate and is never emitted — safer to send unconditionally when `rrn` is absent.

plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
```diff
@@ -119,8 +131,26 @@ def get_investigations(
...
+            # Track all RRNs from current poll for state management (enables deduplication across restarts)
+            latest_investigations.add(rrn)
+            if rrn not in previous_investigations:
+                self.send_investigation(investigation, rrn)
+            else:
+                self.logger.info(f"Duplicate investigation found and skipped: {rrn}")
+        # make sure we don't return an empty set if no new investigations were found, so we don't lose the previous state
```
lcwiklinski-r7: This `else` is unreachable — the method is only called from the `if investigations:` block above, and the loop always adds at least one RRN, so `latest_investigations` is never empty here. It can collapse to `return latest_investigations`, and the "no new investigations" wording is misleading since the set holds *all* retrieved RRNs, duplicates included.

plugins/rapid7_insightidr/plugin.spec.yaml
```diff
@@ -36,6 +36,7 @@ sdk:
   version: 6.6.0
   user: nobody
 version_history:
+  - "12.0.10 - Trigger `Get New Investigations` updated for IDR API latency issues and dedupe logic added | Updated depdancies"
```
lcwiklinski-r7: Typo: `depdancies` → `dependencies` (prior entries spell it correctly), and since this mirrors verbatim into `help.md`, fix it here and regenerate. While here, consider matching the existing trigger-entry style (cf. 12.0.2), e.g. `` Triggers: `Get New Investigations` - Overlap poll windows for IDR API indexing latency and add RRN deduplication | Updated dependencies``.

plugins/rapid7_insightidr/help.md
```diff
@@ -3428,6 +3428,7 @@ Example output:
 
 # Version History
 
+* 12.0.10 - Trigger `Get New Investigations` updated for IDR API latency issues and dedupe logic added | Updated depdancies
```
lcwiklinski-r7: Same `depdancies` → `dependencies` typo — fix it in `plugin.spec.yaml` and regenerate rather than editing here, or it'll be reintroduced.

plugins/rapid7_insightidr/unit_test/test_get_new_investigations_trigger.py
```diff
@@ -0,0 +1,262 @@
...
+class StopLoop(Exception):
+    """Exception raised from mocked time.sleep to break trigger's while loop"""
+
+
+@patch(
+    "komand_rapid7_insightidr.triggers.get_new_investigations.trigger.GetNewInvestigations.get_current_time",
+    return_value=datetime(2026, 8, 26, 12, 0, 0, tzinfo=UTC),
+)
+@patch("time.sleep", side_effect=StopLoop)
+@patch("komand_rapid7_insightidr.triggers.get_new_investigations.trigger.GetNewInvestigations._call_search_api")
```
lcwiklinski-r7, line 28: The retry/backoff path (the `except` block in `run()`) isn't exercised — worth a case where `_call_search_api` raises, asserting it retries and eventually re-raises after `MAX_NUMBER_OF_RETRIES`.

plugins/rapid7_insightidr/unit_test/test_get_new_investigations_trigger.py
```diff
@@ -0,0 +1,262 @@
...
+    def setUp(self) -> None:
+        self.trigger = GetNewInvestigations()
+        self.trigger.connection = MagicMock()
+        self.trigger.logger = MagicMock()
+        self.trigger.send = MagicMock()
+        self.trigger._save_state = MagicMock()
+        self.trigger.state = {}
+        self.params = {"frequency": 15, "search": []}
+
+    @parameterized.expand(
```
lcwiklinski-r7, line 39: None of these cases cover an investigation missing `rrn`, or a duplicate `rrn` within a single poll — both are plausible responses, and the `"N/A"` default plus previous-poll-only membership check give surprising results. A regression case would be worth adding.

plugins/rapid7_insightidr/unit_test/test_get_new_investigations_trigger.py
```diff
@@ -0,0 +1,262 @@
...
+                [
+                    {"rrn": "rrn:investigation:1", "title": "Investigation 1"},
+                    {"rrn": "rrn:investigation:2", "title": "Investigation 2"},
+                    {"rrn": "rrn:investigation:3", "title": "Investigation 3"},
+                ],
+                2,
+                {"rrn:investigation:1", "rrn:investigation:2", "rrn:investigation:3"},
+            ],
+            [
+                "empty API response preserves state",
```
lcwiklinski-r7, line 73: nit: with `data=[]` the `if investigations:` guard is false, so `run()`'s own `else` runs and dedupe is never called — the name "empty API response preserves state" is a little misleading, and it confirms the `else` in `dedupe_and_send_investigations` is unreachable.

plugins/rapid7_insightidr/unit_test/test_get_new_investigations_trigger.py
```diff
@@ -0,0 +1,262 @@
...
+        _mock_sleep,
+        _mock_get_current_time,
+        _test_name,
+        initial_state,
+        api_response,
+        expected_sends,
+        expected_state,
+    ):
+        self.trigger.state = initial_state.copy()
+        mock_call_search_api.return_value = {"data": api_response, "metadata": {"total_pages": 1}}
```
lcwiklinski-r7, line 114: nit: every mock hardcodes `total_pages: 1`, so the pagination loop in `get_investigations` is never tested — a multi-page case would confirm pages are fetched with the right `index` and concatenated.

plugins/rapid7_insightidr/unit_test/test_get_new_investigations_trigger.py
```diff
@@ -0,0 +1,262 @@
...
+        sent_rrns = [call_args[0][0][Output.INVESTIGATION]["rrn"] for call_args in self.trigger.send.call_args_list]
+        self.assertEqual(
+            sent_rrns, ["rrn:investigation:A", "rrn:investigation:B", "rrn:investigation:C", "rrn:investigation:D"]
+        )
+
+        # Verify final state contains only RRNs from last poll
+        final_state = set(self.trigger.state.get("RRNs", []))
+        self.assertEqual(final_state, {"rrn:investigation:C", "rrn:investigation:D"})
+
+    def test_state_loaded_from_persistence_on_restart(self, mock_call_search_api, mock_sleep, mock_get_current_time):
```
lcwiklinski-r7: This reads as the restart case but only re-checks loaded-state dedup (same path as the "mixed" case). Could we add a case with an already-sent RRN that's *not* in the seeded state, to pin down whether the 5-min state horizon vs 10-min restart lookback re-emits it?

plugins/rapid7_insightidr/unit_test/test_get_new_investigations_trigger.py
```diff
@@ -0,0 +1,262 @@
...
+
+        with self.assertRaises(StopLoop):
+            self.trigger.run(self.params)
+
+        # Verify _call_search_api called with correct time window in payload
+        call_args = mock_call_search_api.call_args
+        payload = call_args[0][3]  # Fourth positional arg is payload (resource_helper, endpoint, method, payload)
+
+        # First poll: start_time should be current_time - 10 minutes (INITIAL_LOOKBACK_MINUTES)
+        expected_start = (mock_now - timedelta(minutes=10)).isoformat()
```
lcwiklinski-r7: nit: this hardcodes `timedelta(minutes=10)` while `test_subsequent_poll_applies_5_minute_overlap` uses the imported `INITIAL_LOOKBACK_MINUTES` — reuse the constant so the test tracks the value automatically.

plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
joneill-r7: I've included a comment when we actually apply this var calculation that it's for the late indexing on line ~86. Looking at the trigger code pre `12.0.2` we used to always do a 20 minute lookback on every run, this comment is to ensure no one increases the `API_LATENCY_OVERLAP_MINUTES` beyond the value of `INITIAL_LOOKBACK_MINUTES`. eg if we set `INITIAL_LOOKBACK_MINUTES` to 20 then this would extend beyond the initial look back? Added an additional unit test to safeguard this in future as well.

plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
joneill-r7: Yeah I wasn't a fan of this default N/A existing in the first place, i think we have bigger problems if the API starts returning investigations with no RRNs, but still a valid reasoning to send these. Addressed in latest commit.

plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
joneill-r7: Updated in latest commit, this is a leftover logger from when I didn't persist investigations duped between runs.

plugins/rapid7_insightidr/unit_test/test_get_new_investigations_trigger.py
joneill-r7: In this case it would have sent the ibvestigation again - I've updated the logic to handle this case and added a new unit test.

plugins/rapid7_insightidr/unit_test/test_get_new_investigations_trigger.py
joneill-r7: This is already covered here: `unit_test/test_resource_helper.py` ?

plugins/rapid7_insightidr/unit_test/test_get_new_investigations_trigger.py
joneill-r7: I'm not entirely sure on the plausible comment around N/A as mentioned on another review comment, but it is worth adding anyway. Added in latest commit.

plugins/rapid7_insightidr/unit_test/test_get_new_investigations_trigger.py
joneill-r7: I've removed the `else` block in the dedupe method, but I think we should keep this to preserve this behaviour in future releases that this logic isn't changed.

plugins/rapid7_insightidr/unit_test/test_get_new_investigations_trigger.py
joneill-r7: added a pagination test, I was trying to keep the changes changes minimal scope in this PR but it will do no harm to add more coverage 😄

igorski-r7 (approved):

plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
```diff
@@ -119,8 +136,26 @@ def get_investigations(
...
         return investigations
 
-    def send_investigation(self, investigation: Dict[str, Any]) -> None:
-        self.logger.info(f"Investigation found: {investigation.get('rrn', 'N/A')}")
+    def dedupe_and_send_investigations(self, investigations: List[Dict[str, Any]], previous_investigations: set) -> set:
+        latest_investigations = set()
+        for i, investigation in enumerate(investigations):
+            if rrn := investigation.get("rrn"):
+                # Track all RRNs from current poll for state management (enables deduplication across restarts)
+                latest_investigations.add(rrn)
```
lmacoun-r7, line 144: are the investigations sorted? can we add investigation to `latest_investigations` set which isn't the most recent one?

plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
```diff
@@ -119,8 +136,26 @@ def get_investigations(
...
+            else:
+                self.logger.warn(
+                    f"Investigation {i}: does not have an RRN, skipping deduplication for this investigation."
```
lmacoun-r7, line 147: is it possible to have investigation without rrn?

plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
joneill-r7: there is pagination else where within the trigger to collates all the investigations before we hit this dedupe logic if this is what you mean? https://github.com/rapid7/insightconnect-plugins/blob/75c4a9632dac6fdc34e38e57fd905d22d207513c/plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py#L129. so this would still be deduping across each trigger iteration.

plugins/rapid7_insightidr/komand_rapid7_insightidr/triggers/get_new_investigations/trigger.py
joneill-r7: I think this is very unlikely, but it was flagged by Lukasz that I was technically allowing for a bug that we'd dedupe if more than one investigation didn't have an RRN here, and thought that we've had many other problems on the IDR API that a simple else with a logger wasn't the worst to include: https://github.com/rapid7/insightconnect-plugins/pull/4049#discussion_r3893532681

lmacoun-r7 (approved):

lcwiklinski-r7 (approved):

PR #4052, opened 2026-08-31
[SOAR-21944] fix(list_group_members): paginate members via @odata.nextLink

lcwiklinski-r7:
List Group Members issued a single GET to groups/{id}/members and returned only the first page (Microsoft Graph caps at 100 members/page) while reading count from @odata.count — the true server-side total. For groups with more than 100 members the returned members and count disagreed, and members past the first 100 were silently dropped.

Follow @odata.nextLink in a bounded loop (mirroring the search_device action) to accumulate every page, keeping count from the first page's @odata.count. Also bumps the SDK 6.5.1 -> 6.6.0 and the plugin version 6.0.0 -> 6.0.1, and adds a regression unit test covering cross-page member accumulation.

Fixes: SOAR-21944

## 🎫 Ticket
https://rapid7.atlassian.net/browse/SI-35043

## 🧩 Type of Change

- [ ] Feature
- [ ✅ ] Bug fix
- [ ] Other

## 🧪 Testing
- Unit tests
- Manual testing steps

### DIFF
--- a/plugins/azure_ad_admin/plugin.spec.yaml
+++ b/plugins/azure_ad_admin/plugin.spec.yaml
@@ -4,16 +4,17 @@ products: [insightconnect]
 name: azure_ad_admin
 title: Microsoft Entra ID Admin
 description: "Microsoft Entra ID Admin performs administrative tasks in Microsoft Entra ID (formerly Azure AD).\n\nIt uses the [User](https://docs.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0) endpoint in the [Microsoft Graph API](https://docs.microsoft.com/en-us/graph/overview?view=graph-rest-1.0)"
-version: 6.0.0
+version: 6.0.1
 connection_version: 4
 vendor: rapid7
 cloud_ready: true
+enable_cache: true
 support: community
 supported_versions: ["2022-05-30"]
 status: []
 sdk:
   type: slim
-  version: 6.5.1
+  version: 6.6.0
   user: nobody
 key_features:
   - Add and remove users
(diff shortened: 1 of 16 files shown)


plugins/azure_ad_admin/plugin.spec.yaml
```diff
@@ -13,7 +13,7 @@ supported_versions: ["2022-05-30"]
 status: []
 sdk:
   type: slim
-  version: 6.5.1
+  version: 6.6.0
```
joneill-r7, line 17: can you double check please that the trigger starts up ok on this image? I think bumping the SDK here introduces the requirement for the `enable_cache` in the spec? i might be wrong, this should be caught by the validators but bug exists here: https://rapid7.atlassian.net/browse/SOAR-21948

joneill-r7: ah this plugin is cloud enabled and that validator did flag actually 😄

lcwiklinski-r7: Yeah, I've been about to push enable cache to this PR. Seems like it starts locally with `insight-plugin run tests/risk_detection.json`, however, it returns 
```
 Getting token from: https://login.microsoftonline.com/xxxxxxx-xxxx-xxxxxx-xxxx-xxxxxx/oauth2/token

2026-08-31 08:47:35 [info     ] Authentication request status: 200

2026-08-31 08:47:35 [info     ] Authentication Token: ****************eNe_g

2026-08-31 08:47:35 [error    ] No trigger ID found in connection meta. Trigger will not be stateful.

2026-08-31 08:47:35 [info     ] rapid7/Microsoft Entra ID Admin:6.0.1. Step name: risk_detection

2026-08-31 08:47:37 [error    ] An error occurred during plugin execution! Risk Detections returned an unexpected response: 403 Please contact support for help. Response was {"error":{"code":"Forbidden","message":"You cannot perform the requested operation, required scopes are missing in the token.
....
```

Do you think I should revert the SDK? @joneill-r7

joneill-r7 (approved):

plugins/azure_ad_admin/plugin.spec.yaml
joneill-r7: latest changes look good to me!

PR #4055, opened 2026-08-31
[SOAR-21907] deps fix for netmiko

lcwiklinski-r7:
## 🎫 Ticket
SOAR-21907

## 🧩 Type of Change

- [ ] Feature
- [ ] Bug fix
- [ ✅  ] Other

### DIFF
--- a/plugins/netmiko/requirements.txt
+++ b/plugins/netmiko/requirements.txt
@@ -1,4 +1,5 @@
 # List third-party dependencies here, separated by newlines. 
 # All dependencies must be version-pinned, eg. requests==1.2.0
 # See: https://pip.pypa.io/en/stable/user_guide/#requirements-files
+netmiko==4.6.0
 paramiko==5.0.0
--- a/plugins/netmiko/.CHECKSUM
+++ b/plugins/netmiko/.CHECKSUM
@@ -1,5 +1,5 @@
 {
-	"spec": "fae78493ed7d87a0fc7f4e98a1f03d97",
+	"spec": "5dae4840e7f4168eced611d5ca90b218",
 	"manifest": "36cfe7a9edc50d20113eb8c7000cf9d8",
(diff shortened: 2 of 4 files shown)


plugins/netmiko/requirements.txt
```diff
@@ -1,4 +1,5 @@
 # List third-party dependencies here, separated by newlines. 
 # All dependencies must be version-pinned, eg. requests==1.2.0
 # See: https://pip.pypa.io/en/stable/user_guide/#requirements-files
+netmiko==4.6.0
```
joneill-r7, line 4: we had netmiko == 3.3.0 before, and then 4.7.0. can we still not have 4.7.0 in requirements.txt like it was moved into dockerfile and then upgrade paramiko after? also i'd be try to see what bumping paramiko 2 majors does under the hood for these requirements?

lcwiklinski-r7: @joneill-r7 

https://github.com/ktbyers/netmiko/pull/3852#issuecomment-4442806949

https://github.com/ktbyers/netmiko/blob/v4.7.0/pyproject.toml#L27

https://github.com/ktbyers/netmiko/releases/tag/v4.7.0
Their release log says `Update poetry lock file to pin Paramiko version <5 (paramiko5 caused significant breakage) in https://github.com/ktbyers/netmiko/pull/3852`

cc: @igorski-r7

joneill-r7: ok so I'm guessing this means it will cause problems for the plugin?

lcwiklinski-r7: yes @joneill-r7 this url explains it mostly https://github.com/ktbyers/netmiko/pull/3852#issuecomment-4442806949

igorski-r7 (approved):
LGTM 😃

PR #4061, opened 2026-09-07
[SOAR-22002] Get incident comments worknotes

lcwiklinski-r7:
## 🎫 Ticket
https://rapid7.atlassian.net/browse/SOAR-22002

## 🧩 Type of Change

- [ ] Feature
- [ ✅ ] Bug fix
- [ ] Other

 ## 🧠 Background & Motivation

  SI-35318: `Get Incident Comments and Work Notes` returned an empty list with no error,
  even though the incident visibly had comments and work notes in the ServiceNow UI.

  The action reads the `sys_journal_field` table, which enforces read Access Controls of
  its own, independently of the `incident` table. When those Access Controls filter the
  connected account out, ServiceNow does not return an error — it returns `HTTP 200` with
  `{"result": []}`. The plugin therefore reported success with no data, and nothing in the
  connection pointed at a permissions problem: `Test Connection` passes because it queries
  `cmdb_ci`, not the journal table.

  Confirmed on an internal instance: an account that reads incidents fine gets zero rows
  from `sys_journal_field` unless a role it holds is added to that table's `read` Access
  Control.

## ✨ What Changed

`Get Incident Comments and Work Notes` now falls back to reading the journal off the
incident record when the `sys_journal_field` query returns nothing:

- `GET /api/now/table/incident/<sys_id>?sysparm_fields=comments,work_notes&sysparm_display_value=true`.
  `sysparm_display_value=true` is required — read as a raw value, a journal field holds
  only its most recent entry.
- New `icon_servicenow/util/journal_helper.py` parses the display value into the same
  output objects the `sys_journal_field` query produces, so the action's output shape is
  unchanged. Header parsing is structural, not format-based: the timestamp follows the
  caller's date format and time zone and the field label is renameable and localizable.
  Text whose header is not recognised is still returned, with an empty author and
  creation date, rather than dropped.
- The `Type` input is respected in the fallback (`all` / `comments` / `work notes`).
- The fallback never turns a working call into an error: if the incident cannot be read
  either, it logs the reason and returns no entries.
- Instances where the account can read `sys_journal_field` are untouched — the fallback
  only runs on an empty result, and a unit test asserts that the readable-journal path
  still makes exactly one HTTP call.

Two limitations of the fallback, documented under Troubleshooting: entries have an empty
`sys_id` and `sys_tags`, and `sys_created_by` is the author's display name rather than
their user name. Neither is available on the incident record.

Docs: the Requirements list now names `sys_journal_field` for this action and states the
remedy — a `security_admin` adds a role the account already holds, such as `itil`, under
`Requires role` on that table's `read` Access Control. Granting `admin` is not required.

Also included: SDK bumped to 6.6.0, and `enable_cache: true`, which the current validator
requires for a cloud-ready plugin with triggers (a pre-existing gap, unrelated to the fix).

## 🧪 Testing

**Unit tests** — 38 new tests, 63 passing in the plugin suite.

- Fallback for each `Type` value, plus the readable-journal path asserting output is
  unchanged and only one request is made.
- Incident not readable (404), unexpected response shape, invalid system ID, no comments
  or work notes, and a journal query rejected outright (403) which must still raise.
- 22 parser cases: multi-paragraph entries, CRLF, a localized field label, 12-hour and
  month-name timestamps, an author containing ` - ` or parentheses, an author with no last
  name, and six cases of entry text that looks like a header (`Called at 10:15:00 - no
  answer (voicemail)`) which must stay entry text.
- A ReDoS regression test: the header pattern's repetitions are bounded, and unbinding
  them takes matching from 0.1 ms to over 15 s on a single pasted-log line.

**Manual testing** — internal ServiceNow instance, in the exact SI-35318 state (the
account reads the incident, `sys_journal_field` returns zero rows):

- `all`, `comments` and `work notes` all returned the entries; the per-type sums add up
  to `all`.
- Same incident and account against the 8.1.5 image returns an empty list, against this
  build returns the text.
- With the role added to the table's `read` Access Control, entries come from
  `sys_journal_field` again with `sys_id` populated, confirming the documented remedy.

`insight-plugin validate` passes.
  `Get Incident Comments and Work Notes` now falls back to reading the journal off the
  incident record when the `sys_journal_field` query returns nothing:

  - `GET /api/now/table/incident/<sys_id>?sysparm_fields=comments,work_notes&sysparm_display_value=true`.
    `sysparm_display_value=true` is required — read as a raw value, a journal field holds
    only its most recent entry.
  - New `icon_servicenow/util/journal_helper.py` parses the display value into the same
    output objects the `sys_journal_field` query produces, so the action's output shape is
    unchanged. Header parsing is structural, not format-based: the timestamp follows the
    caller's date format and time zone and the field label is renameable and localizable.
    Text whose header is not recognised is still returned, with an empty author and
    creation date, rather than dropped.
  - The `Type` input is respected in the fallback (`all` / `comments` / `work notes`).
  - The fallback never turns a working call into an error: if the incident cannot be read
    either, it logs the reason and returns no entries.
  - Instances where the account can read `sys_journal_field` are untouched — the fallback


igorski-r7 (commented):
Went through the fallback path end to end. Logic is safe, it never turns a working call into an error, it respects the `type` input, and the existing `sys_journal_field` path stays untouched. Traced the error handling through `request_helper.make_request` to confirm the `404`, non JSON, and `403` cases all degrade the way you describe.

Left a couple of small comments below, **nothing blocking**.

plugins/servicenow/icon_servicenow/util/journal_helper.py
```diff
@@ -0,0 +1,139 @@
...
+    :param logger: Logger object of the running action
+    :type logger: Logger
+    :param system_id: System ID of the incident to read comments and work notes from
+    :type system_id: str
+    :param entry_type: Either 'all', 'comments' or 'work_notes'
+    :type entry_type: str
+    :return: List of comment and work note objects
+    :rtype: list
+    """
+    if not (system_id or "").isalnum():
```
igorski-r7: Nice catch gating `system_id` with `.isalnum()` before dropping it into the incident URL here, that blocks path or query injection. The existing `sys_journal_field` query build in `action.py` does the same string interpolation **without** that guard. Not asking for a fix in this PR since it is pre-existing and out of scope, just might be worth a quick follow up ticket.

plugins/servicenow/icon_servicenow/util/journal_helper.py
```diff
@@ -0,0 +1,139 @@
...
+
+from insightconnect_plugin_runtime.exceptions import PluginException
+from requests import RequestException
+
+# A journal field read with sysparm_display_value=true holds every entry of that field in one string,
+# each a header line - '2019-09-26 21:19:11 - System Administrator (Work notes)' - followed by its text,
+# which may itself contain blank lines. The timestamp honours the caller's date format and time zone and
+# the label is the field's own label, so headers are matched by shape rather than against a fixed format,
+# with every repetition bounded to keep matching linear on the pasted logs that end up in comments.
+journal_entry_header = re.compile(
```
igorski-r7, line 16: Looking at the `renamed_label_containing_parentheses` case, if a field label itself contains parens like `Comments (public)`, the author ends up captured as `Joe Employee (Comments`. Already tested and called **best effort** in the docstring, so probably known, just want to double check this is the tradeoff you want to accept rather than something to tighten up.

igorski-r7:
Test coverage looks solid. **38 cases**, the ReDoS regression test, non JSON body handling, and the suppressed field case. Did not find any gaps.

plugins/servicenow/icon_servicenow/util/journal_helper.py
lcwiklinski-r7: Yes, accepted tradeoff: `Joe Employee (Contractor) (Work notes)` needs the last parentheses as the label, `Joe Employee (Comments (public))` needs the first, and the two are indistinguishable to the parser — so I kept the case that's common in ServiceNow. Only `sys_created_by` is affected; the text, timestamp and entry splitting are all correct.

plugins/servicenow/icon_servicenow/util/journal_helper.py
lcwiklinski-r7: Good catch, filing the follow-up — 15 call sites in the plugin interpolate an input into a URL and this guard is the only validation in `icon_servicenow/`, including `delete_incident` and `read_incident` which build a path. Too broad for a bugfix PR; I'll post the ticket key here.

igorski-r7 (approved):

landrisek-r7 (approved):

